PR #6, opened 2012-01-31
Fix for issue 51.

lemval:
Changed JGitUtils.java to not traverse unaccessible subfolders (issue 51).
This by checking if the folder is readable for the user running the gitblit server.
If the folder isn't readable, there is no sense in progressing further down the line.


andrm:
I really would need a release with PR #6 init....

PR #162, opened 2014-03-22
LDAP: Authenticated Searches without a manager password

j3rem1e:
Allow to use the LDAP AuthProvider with a LDAP Server
prohibiting anonymous searches but without providing
a manager password : searches are made on behalf of
the authenticated user.

### DIFF
--- a/src/main/distrib/data/gitblit.properties
+++ b/src/main/distrib/data/gitblit.properties
@@ -1516,6 +1516,15 @@ realm.ldap.username = cn=Directory Manager
 # SINCE 1.0.0
 realm.ldap.password = password
 
+# Bind pattern for Authentication.
+#
+# Allow to authenticate an user without LDAP Searches.
+# 
+# e.g. CN=${username},OU=Users,OU=UserControl,OU=MyOrganization,DC=MyDomain
+#
+realm.ldap.bindpattern = 
+
+
 # Delegate team membership control to LDAP.
 #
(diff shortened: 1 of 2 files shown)


fzs:
I need to ask: what problem does this change try to solve? Since Gitblit will always bind as the manager first, a manager password will always have to be provided. What is the intented benefit of searching for the user account as the user instead of the manager? That should render the manager account obsolete, but the commit doesn't actually do that.

j3rem1e:
When this patch was proposed, the manager password was not mandatory. you can provide an empty user/password. I don't know however if something has changed since v1.6

This patch allow to use an ldap server without the manager password, when the server doesn't allow anonymous search : it binds as the user. I use it every day with my gitblit instance with a corporate active directory.

fzs:
So you use this with an empty username and password or just an empty password? Does your manager account exist and have an empty password?

I haven't understood yet how this works. In your version, too, `authenticate` will first call `getLdapConnection`. Only if it gets back a connection will it run your new code. But `getLdapConnection` already binds either anonymously (empty username and empty password) or with the username. So if there is a username to bind against but no password, and the bind thus fails, `authenticate` will stop.

fzs:
Let me ask differently: Do you expect that, with this change, when the `bindpattern` property is not empty, the `username` and `password` properties are ignored, and Gitblit never binds as the manager but only as the user?

PR #203, opened 2014-06-18
Add setting to make the Git repo URL configurable

svogt:
When Gerrit and Gitblit are deployed on the same server /r/ might be used by Gerrit. Add a config setting to change the shown URL to another primary URL


gitblit:
@svogt did you still need this functionality?  While it is true that both Gitblit and Gerrit use `/r/` they both can't be mounted on the same context path so I'm not sure if the motivation in the description is enough to justify adding this setting.

PR #214, opened 2014-08-12
Extract user name from client certificate with regex pattern.

cdroescher:
Introduce a property "server.regexClientCert" and "server.clientCertUserToUppercase" to have the opportunity to extract the username with a self defined regex pattern and use the username to perform a login.


gitblit:
Closing incomplete PR.

PR #230, opened 2014-11-17
IE < 10 does not support gradients.

tomaswolf:
Now this is embarrassing. I did after all find an old PC on which to test this with IE 8 and got a crappy UI missing most but not all parts. Quickly discovered that [IE < 10 does not support gradients](http://caniuse.com/css-gradients), which I had used extensively. Had never occurred to me that even IE9 might not have this.

Since IE9 and IE8 are still in use (a total of 5% according to caniuse, some other sites report much higher values) I guess it's worth catering to users who have to use such old browsers (especially corporate users who have an unfortunate tendency to lag behind).

See the commit comment for the changes. Basically, I've replaced all gradients by simpler CSS or PNGs. Visually the UI is pretty much the same, and it now works (and looks good) even on poor old IE8.

### DIFF
--- a/src/main/resources/gitblit.css
+++ b/src/main/resources/gitblit.css
@@ -1440,6 +1440,8 @@ div.diff > table {
 
 /* Image diffs. */
 
+/* Note: can't use gradients; IE < 10 doesn't support them. Use pre-created pngs with transparency instead. */
+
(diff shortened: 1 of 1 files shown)


fhtthg:
some thing i test in my win10 it ti's work so happy

PR #242, opened 2015-01-15
Change Jenkins script so it works when pushing/pulling over SSH

rdicroce:
The Jenkins post-receive hook doesn't work if you push changesets over SSH (url variable is only populated for HTTP/HTTPS pushes) and it also doesn't work if Jenkins is configured to pull changesets over SSH (the Jenkins Git plugin matches by host + path and the SSH URL has a different path compared to the HTTP/HTTPS URL).

This change resolves both issues.

### DIFF
--- a/src/main/distrib/data/groovy/jenkins.groovy
+++ b/src/main/distrib/data/groovy/jenkins.groovy
@@ -67,10 +67,17 @@ logger.info("jenkins hook triggered by ${user.username} for ${repository.name}")
 
 // define your jenkins url here or set groovy.jenkinsServer in 
 // gitblit.properties or web.xml
+// you can also set the GitBlit base url to use
+// it is needed if you push changes using SSH (url variable is not populated in that case)
 def jenkinsUrl = gitblit.getString('groovy.jenkinsServer', 'http://yourserver/jenkins')
+def gitblitUrl = gitblit.getString('groovy.jenkinsGitBlitHost', url)
 
-// define the trigger url
-def triggerUrl = jenkinsUrl + "/git/notifyCommit?url=${url}/r/${repository.name}"
+// the host and path must match for the Git plugin to trigger a pull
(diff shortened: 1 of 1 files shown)


gitblit:
I'm going to solve this from within Gitblit.  Gitblit should be exporting the appropriate URL and it is not.

rdicroce:
Exporting the appropriate URL will fix the problem when a user pushes over SSH, but not the problem when Jenkins is configured to pull over SSH. Either the SSH URL would have to be changed to match the HTTP/HTTPS URL (by adding the "/r" part) or the Jenkins script would have to be changed as I have done so that it calls both variants.

gitblit:
I should have been more clear.  Gitblit should be exporting the complete repository url, not just a base url.

rdicroce:
That would also work, but GitBlit would need to export the complete URL for every enabled transport, and the script would need to call all of those URLs. Otherwise the hook won't work if the push occurs with a different transport than than Jenkins is configured with (e.g. developer pushes over HTTP, Jenkins pulls over SSH).

gitblit:
Hmm.  Perhaps Jenkins should trigger on a repo name and not a url.  Are the alternatives?

rdicroce:
We should maybe discuss with the author of the Jenkins Git plugin. From looking at the code, it actually doesn't match the entire URL, just the host and the path. That's enough to cause a problem with GitBlit because the SSH path is different from the HTTP/HTTPS path.

Jenkins is capable of triggering builds using tokens (this can be done with any source control, it's not part of the Git plugin) but then the user has to explicitly configure a project-specific token on both Jenkins and GitBlit. Not a huge problem I suppose, but it is an extra step and the user has to make sure the token is unique for every project.

gitblit:
The recently merged #986 may offer an alternative solution for this issue.  Closing for now.

PR #243, opened 2015-01-20
Applying commercial certificate to gitblit

iloving:
1. Get the gitblit keystore password from /opt/gitblit/data/gitblit.properties
2. There are two keys in the keystore. localhost, and one other. Delete the other.  Use Portacle for this to make it easier.
3. keytool -importkeystore -srckeystore <PKCS12 file> -srcstoretype PKCS12 -destkeystore serverKeyStore.jks
4. Enter the passwords for the key store and for the certificate
5. Important: change the password of the imported certificate so that it is the same as the keystore.  Gitblit uses the same configured password for both.
6. Gitblit uses the first key it finds in the keystore, which should be the one we just imported. If not, you need to get the alias from the keystore and update server.certificateAlias in gitblit.properties
7.Restart gitblit


gitblit:
@iloving Thanks for the write-up.  Closing this PR as there is nothing to merge.

PR #244, opened 2015-02-15
Fix spelling mistake

Southparkfan:
Fixed a spelling mistake here, though I wonder if that's enough...

If I would create a tag in gitblit now, you would see "[author] nieuwe tag gemaakt [tag] in [repository]". I wonder if "[author] heeft een nieuwe tag gemaakt ([tag]) in [repository]" isn't better - but since I'm not familiar with i18n in gitblit, I'll leave that up to you.


kwoot:
Dank!

kwoot:
I mean: Thanks!

gitblit:
Merged.

PR #246, opened 2015-03-06
Proposal for fix of http performance problem

pierreyvesbaloche:
Hi James,

I'm a coworker of Johann Ollivier Lapeyre (from the discussion https://groups.google.com/forum/#!topic/gitblit/eEGZpiApDTQ) and I have the following patch to suggest.
I tried it locally and it moved in our test environnent from 11s down to 2s for a user connection.

Best regards

Pierre-yves Baloche

### DIFF
--- a/src/main/java/com/gitblit/auth/AuthenticationProvider.java
+++ b/src/main/java/com/gitblit/auth/AuthenticationProvider.java
@@ -15,7 +15,13 @@
  */
 package com.gitblit.auth;
 
+import java.io.ByteArrayOutputStream;
 import java.io.File;
+import java.io.IOException;
+import java.io.ObjectOutputStream;
+import java.math.BigInteger;
+import java.security.MessageDigest;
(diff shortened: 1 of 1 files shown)


gitblit:
Looks great.  Can we swallow the exceptions in `checksum` and not bubble them up?  All Java implementations are required to have an SHA1 digest implementation so those exceptions should never trigger.

gitblit:
I adjusted your contribution a little.  Thanks!

pierreyvesbaloche:
Glad I could help ;-) Of course your modifications were well appropriated. Cheers.

Benjamin321:
Thank you for this patch. With about 30 active (of 100) users and more then 1500 repositories, push and pull sometimes needed more then a minute to complete. Regular time for requests was around 30 seconds. 
In the monitoring I saw a lot of read/write traffic on the users.conf. After upgrading to 1.7.1 traffic has become way less and push/pull requests rarely need more then five seconds.

PR #247, opened 2015-03-23
(ticket-537) user bindUser for group detection

RainerW:
This should is a fix for *( https://code.google.com/p/gitblit/issues/detail?id=537) *
-> For security reason a normal user cannot read groups from our LDAP, so in this case the search user has to be used for group/team detection.

would appreciate some feedback : 
 1: Based on the comment "// Binding will stop any LDAP-Injection Attacks ...." should there be a separate setting for enabling this 'feature', or what actually is a LDAP-Injection Attack?
 2: getLdapConnection() should  now be called createLdapConnection() ?

**Edit:** 
*Original google code ticket is not available anymore?*
Problem was:
 - inital ldap bind does work to authenticate the user
 - the new connection was then used to no query the team memberships
   -  his works with an AD, but on OpenLDAP a user has not the right's to see the groups
 - so in (our) OpenLDAP setup the Team memberships have to user the 'old' non user ldap connection

### DIFF
--- a/src/main/java/com/gitblit/auth/LdapAuthProvider.java
+++ b/src/main/java/com/gitblit/auth/LdapAuthProvider.java
@@ -321,7 +321,6 @@ public class LdapAuthProvider extends UsernamePasswordAuthenticationProvider {
 				if (result != null && result.getEntryCount() == 1) {
 					SearchResultEntry loggingInUser = result.getSearchEntries().get(0);
 					String loggingInUserDN = loggingInUser.getDN();
-
 					if (alreadyAuthenticated || isAuthenticated(ldapConnection, loggingInUserDN, new String(password))) {
 						logger.debug("LDAP authenticated: " + username);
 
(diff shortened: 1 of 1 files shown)


gitblit:
At a quick glance you are changing the existing behavior to solve your issue, but you may be breaking other installs.  This needs some discussion on the mailing list.  Clearly with LDAP what works for some may not work for all - hence the reason for your PR.

RainerW:
created new branch and extracted new behaviour into config key

RainerW:
rebased to develop, not sure about the "since 1.7.0" Information in defaults.properties

gitblit:
Hi @RainerW.

Github doesn't allow you to merge a PR to another branch AFAIK i.e. if I merge now, your change will merge to `master` even though you rebased on `develop` because `master` is the branch for which the PR was initially opened.

Having said that, while your change solves your problem and may solve #920, I don't like this solution - changing the behavior of `isAuthenticated`.  If the problem is related to group/team search then the fix should be isolated to the connection used in `getTeamsFromLdap`.

PR #249, opened 2015-03-24
Docs: Fix RPC URL

StephenKing:
The RPC URL should be `/rpc/?` and not `/rpc?` according to
https://groups.google.com/d/msg/gitblit/Ajp2gR3B2bM/AXg6wIf21eIJ


StephenKing:
Uff.. should create change based on `develop` branch. Will use command line now.. weird web interface

PR #251, opened 2015-04-12
GO: Unexpected disclosure of binary code / Addition of a resource guard filter

jibee:
Hi,

The GO setup allows to download GitBlit's java classes (.class files), and also exposes the whole class structure. A live example can be found there (URL broken on purpose to prevent indexation): https://dev.gitblit.com/ com/gitblit/GitBlit.class

The JAR containing the application and the WAR injected into Jetty are the same file. However Jetty expects to serve the entire WAR contents, except the WEB-INF folder. Thus, all java binary classes in the JAR are served as is they were legitimate resources.

This pull request prevents accidental access to 'resources' such as GitBlit java classes by adding a filter to deny them.

### DIFF
--- a/src/main/java/WEB-INF/web.xml
+++ b/src/main/java/WEB-INF/web.xml
@@ -357,5 +357,17 @@
         <filter-name>wicketFilter</filter-name>
         <url-pattern>/*</url-pattern>
     </filter-mapping>
-    
+    <!-- Resource Guard filter -->
+    <filter>
+	<filter-name>resourceGuard</filter-name>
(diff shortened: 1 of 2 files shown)


gitblit:
Good catch.  The fix for this will have to be implemented against develop, not master, which has almost completely eliminated web.xml.  I'd also like to find a solution that does not require adding a servlet filter to the entire Gitblit app; since this is a GO-only problem there should be a GO-only solution.

jibee:
Hi James,

Thanks for your comments; I will close this pull request and create a new one with a neater fix against develop.

PR #254, opened 2015-05-04
Adding Kerberos5/GSS authentication to ssh

fbacchella:
Adding the possibility to define authentication method order for ssh.

The test activates kerberos authentication, but don't use it, as it needs a kdc and keytabs. It plays with authentication method order to ensure that they are used as they should.


gitblit:
It doesn't look like your unit test changes demonstrate/test working GSS/Kerberos5 authentication. Can you create a test config?

fbacchella:
No that part is quite hard: it needs both a keytab for the client and the server. I can do one for the server, but I'm not sure how it can work client side, and without a running kdc. And then it needs to be loaded inside the ssh client. I need to check if the java ssh client provided can do that.

PR #259, opened 2015-05-19
Implemented filtering by repository on my tickets page

jeyoung:
OK, I think I got it right now. Pulled develop, then merged my ticket-57 changes on top of it.


src/main/java/com/gitblit/wicket/pages/MyTicketsPage.java
```diff
@@ -251,12 +266,39 @@ public void populateItem(final Item<Status> item) {
...
+				PageParameters params = queryParameters(queryParam, milestoneParam, statiiParam, assignedToParam, ts.sortBy, ts.desc, repository, 1);
 				item.add(new LinkPanel("sortLink", null, ts.name, MyTicketsPage.class, params).setRenderBodyOnly(true));
 			}
 		};
 		add(sortMenu);
 
+        // by repository
+		List<RepositoryModel> repositoryChoices = getRepositoryModels();
+        RepositoryModel noneChoice = new RepositoryModel();
+        noneChoice.name = Translation.get("gb.all");
```
gitblit: We can't  use `Transation`.  Instead do this:

```java
getString("gb.all");
```

src/main/java/com/gitblit/wicket/pages/MyTicketsPage.java
```diff
@@ -278,6 +320,12 @@ public void populateItem(final Item<TicketSort> item) {
 			}
 		}
 
+        if (!qb.containsField(Lucene.repository.name()) && !StringUtils.isEmpty(repository)) {
```
gitblit: Resetting back to *all* doesn't work right.

If you add something like this it will:
```java
        if (noneChoice != currentRepository && !qb.containsField(Lucene.repository.name()) ...
```

src/main/java/com/gitblit/wicket/pages/MyTicketsPage.java
```diff
@@ -74,6 +76,7 @@ public MyTicketsPage(PageParameters params)	{
 		final String queryParam = (params == null || StringUtils.isEmpty(params.getString("q", null))) ? "watchedby:" + username : params.getString("q", null);
 		final String searchParam = (params == null) ? "" : params.getString("s", null);
 		final String sortBy = (params == null) ? "" : Lucene.fromString(params.getString("sort", Lucene.created.name())).name();
+		final String repository = (params == null) ? "" : params.getString("repository", null);
```
gitblit: Should use `Lucene.repository.name()` for consistency.

gitblit:
Aside from the little comments I made we have two larger problems.  One is easily fixable.  The other.... not sure.

1. You can't search by the repository name.  It's problematic because of punctuation marks, slashes, etc.  Instead we must specify and search for the `rid` which is a hash of the repository name.  So everywhere you specify a repository name value it must be the `rid` instead.  RepositoryModel.getRID() will give you this value.

2. We have a scaling problem.  I have 50+ repos in my test environment.  This generates a lengthy dropdown menu which scrolls offscreen.  And it really won't scale for those with hundreds or thousands of repos so we need a better plan.  The bootstap in-use is really old and we don't have scrollable dropdown menus.  :(  I think we need to list only the repos that are in the infiltered *My Tickets* resultset.  That means we need to execute two Lucene queries.  Hopefully the performance impact will be minor compared to the entire page-loading cycle.

jeyoung:
@James, re-submitted with changes based on your code review. Also changed the repository options to constrain to those relevant to the user's watched tickets.

(Sorry about the changes of tabs to spaces, please ignore white-space differences when you compare.)

src/main/java/com/gitblit/wicket/pages/MyTicketsPage.java
```diff
@@ -52,355 +57,442 @@
...
+        add(new Label("currentRepository", currentRepository == null ? "none" : currentRepository.toString()));
+
+        ListDataProvider<RepositoryModel> repositoryChoicesDp = new ListDataProvider<RepositoryModel>(repositoryChoices);
+        DataView<RepositoryModel> repositoryMenu = new DataView<RepositoryModel>("repository", repositoryChoicesDp) {
+            private static final long serialVersionUID = 1L;
+
+            @Override
+            public void populateItem(final Item<RepositoryModel> item) {
+                final RepositoryModel r = item.getModelObject();
+                PageParameters params = queryParameters(queryParam, milestoneParam, statiiParam, assignedToParam, sortBy, desc, r.getRID(), 1);
```
gitblit:
```java
String rid = r == noneChoice ? null : r.getRID();
```

Then use rid in the `queryParameters` call.

src/main/java/com/gitblit/wicket/pages/MyTicketsPage.java
```diff
@@ -52,355 +57,442 @@
...
+                item.add(new LinkPanel("sortLink", null, ts.name, MyTicketsPage.class, params).setRenderBodyOnly(true));
+            }
+        };
+        add(sortMenu);
+
+        // by repository
+        final List<QueryResult> tickets =
+            query(initializeQueryBuilder(defaultQueryParam, username), 1, Integer.MAX_VALUE, sortBy, desc);
+        final List<RepositoryModel> repositoryChoices = correspondingRepositories(tickets);
+        Collections.sort(repositoryChoices, new Comparator<RepositoryModel>() {
```
gitblit:
```java
Collections.sort(repositoryChoices);
```

RepositoryModel is already `Comparable`.

src/main/java/com/gitblit/wicket/pages/MyTicketsPage.java
```diff
@@ -52,355 +57,442 @@
...
+                final TicketSort ts = item.getModelObject();
+                PageParameters params = queryParameters(queryParam, milestoneParam, statiiParam, assignedToParam, ts.sortBy, ts.desc, repositoryId, 1);
+                item.add(new LinkPanel("sortLink", null, ts.name, MyTicketsPage.class, params).setRenderBodyOnly(true));
+            }
+        };
+        add(sortMenu);
+
+        // by repository
+        final List<QueryResult> tickets =
+            query(initializeQueryBuilder(defaultQueryParam, username), 1, Integer.MAX_VALUE, sortBy, desc);
```
gitblit: query(initializeQueryBuilder(**Lucene.watchedby.matches(username)**, username), 1, Integer.MAX_VALUE, sortBy, desc);

src/main/java/com/gitblit/wicket/pages/MyTicketsPage.java
```diff
@@ -52,355 +57,442 @@
...
+
+        UserModel currentUser = GitBlitWebSession.get().getUser();
+        if (currentUser == null || UserModel.ANONYMOUS.equals(currentUser)) {
+            setRedirect(true);
+            setResponsePage(getApplication().getHomePage());
+            return;
+        }
+
+        final String username = currentUser.getName();
+        final String defaultQueryParam = "watchedby:"+username;
```
gitblit: Eliminate.

src/main/java/com/gitblit/wicket/pages/MyTicketsPage.java
```diff
@@ -52,355 +57,442 @@
...
+
+        final String[] statiiParam = (params == null) ? TicketsUI.openStatii : params.getStringArray(Lucene.status.name());
+        final String assignedToParam = (params == null) ? "" : params.getString(Lucene.responsible.name(), null);
+        final String milestoneParam = (params == null) ? "" : params.getString(Lucene.milestone.name(), null);
+        final String queryParam = (params == null || StringUtils.isEmpty(params.getString("q", null))) ? defaultQueryParam : params.getString("q", null);
```
gitblit:
```java
final String queryParam = (params == null) ? "" : params.getString("q");
```

Gitblit is primarily for those who do, not those who watch; I want the default query to be for developers.

gitblit:
Good work.  We are very close now and this will be a nice usability improvement.

src/main/java/com/gitblit/wicket/pages/MyTicketsPage.java
jeyoung: In this case, I think mentions:<current user> should be part of the default filter.

jeyoung: BTW, there seems to be a bug in the mention detection. The mention is added to the list only if @username appears in the middle of the comment. If the comment starts with @username, it is not detected. Looks like it can be fixed with tweaking the regex.

Also, shouldn't the main ticket body also be processed for mentions instead of just comments?

jeyoung:
@gitblit These changes should be it. If no more review comments, I'll proceed with fixing the mentions bug, err, mentioned above.

gitblit:
Thanks Eddy!  Merged.

jeyoung:
A pleasure. I'd really like to look into implementing tickets across groups. Is that on the roadmap for 1.7.0? Otherwise, I can pick another 1.7.0 item.

gitblit:
Tickets are intricately tied to repositories.  It would take some heroic changes (i.e. likely not mergeable) to accommodate  that.

If you really wanted to explore that I would suggest implementing some sort of ticket-container object with it's own persistence mechanism and UI.  Maybe this could be a plugin but I haven't spent more than 30 seconds thinking about this so maybe that is a dumb idea.  ;)

jeyoung:
Hmm. I forgot about the integration of tickets within repositories, but your idea of a “repository group ticket service” is worth pursuing. I’ll think about that a bit more this week-end.

PR #261, opened 2015-05-22
A bug with Tomcat and SSO

fbacchella:
The way wicket change session is not tomcat friendly and you use it in com.gitblit.wicket.pages.SessionPage. I think it change the session id too late, the reply header have been already sent. It break CAS (the SSO solution we use) as it really on session id to follow user. But changing session should be kept for security reasons. Tomcat can do that on it's own. So I added a option that allows to disable this in case some one might need it.

### DIFF
--- a/src/main/java/com/gitblit/wicket/pages/SessionPage.java
+++ b/src/main/java/com/gitblit/wicket/pages/SessionPage.java
@@ -96,7 +96,9 @@ public abstract class SessionPage extends WebPage {
 					.getAttribute(Constants.AUTHENTICATION_TYPE);
 
 			// issue 62: fix session fixation vulnerability
-			session.replaceSession();
+			if(app().settings().getBoolean(Keys.web.rewriteSession, true)) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/gitblit/wicket/pages/SessionPage.java
```diff
@@ -96,7 +96,9 @@ private void login() {
 					.getAttribute(Constants.AUTHENTICATION_TYPE);
 
 			// issue 62: fix session fixation vulnerability
-			session.replaceSession();
+			if(app().settings().getBoolean(Keys.web.rewriteSession, true)) {
```
gitblit: I won't merge this PR.  But I understand your issue and I propose the following alternative.
```java
if (AuthenticationType.CONTAINER != authenticationType) {
    session.replaceSession();
}
```

If that's acceptable, open a PR with that change and I'll merge it.

fbacchella:
That look fine to me. I'll do that.

PR #262, opened 2015-05-23
Keepsession

fbacchella:
The way wicket change session is not tomcat friendly and you use it in com.gitblit.wicket.pages.SessionPage. I think it change the session id too late, the reply header have been already sent. It break CAS (the SSO solution we use) as it really on session id to follow user. But changing session should be kept for security reasons. Tomcat can do that on it's own. So I added a option that allows to disable this in case some one might need it.

This time done without a specific option.


gitblit:
Do me a favor and squash these into one commit.

gitblit:
If you aren't aware, you can force push PRs on Github so a new PR is not necessary.

fbacchella:
Git is a little black magic for me, but I think I have done what you want.

gitblit:
Perfect.  Thanks!

PR #265, opened 2015-05-26
Invalid kerberos patches, works now and with a test.

fbacchella:


gitblit:
- Need Apache license headers for all source files.
- Need to add the new unit test class to `GitblitSuite`.

fbacchella:
Done

PR #267, opened 2015-06-03
First draft for a customized avatar image

fbacchella:
This is a draft for custom avatars.

This allow to use other source of avatar than Gravatar.

I put some code in GitblitWicketApp, I'm not sur it's the right place to do that.

I tried to write some test, but I don't have the slightest idea where can I look to generate mock object.


gitblit:
Hi Fabrice,

It will take me a couple days to make time to review this.
My first thought is we should find a way to Inject this.

-J

fbacchella:
I look at guice. Now I start to grasp what all the `@Inject` annotations are.

So I added src/test/java/com/gitblit/tests/GravatarTest.java, but it fails with:

	org.apache.wicket.WicketRuntimeException: Use Application.init() method for configuring your application object
		at org.apache.wicket.Application.getSettings(Application.java:921)
		at org.apache.wicket.Application.getResourceSettings(Application.java:649)
		at com.gitblit.wicket.GitBlitWebApp.init(GitBlitWebApp.java:188)
		at com.gitblit.tests.GravatarTest.oneTest(GravatarTest.java:77)

Any idea to work around that ?

gitblit:
Hi Fabrice,

I merged your work and re-implemented it using Guice.  You should now be able to provide your own implementation of `AvatarGenerator` something like:

```java
@Singleton
public class MyGenerator implements AvatarGenerator {

    private final IStoredSettings settings;

    @Inject
    public MyGenerator(IStoredSettings settings) {
        this.settings = settings;
    }

    @Override
    public String getURL(String username, String emailaddress, boolean identicon, int width) {
        return "test";
    }
}
```

I also revised the unit test so it should be more clear how you can test your own implementation.

PR #268, opened 2015-06-04
Support for different post/host in the ssh://... url in the web interface.

bogeskov:
I've tried to run gitblit in a docker container, and wanted to run the ssh server on what looks like port 22.
However I didn't want to run it as root. So I exposed the default port as (server's 2nd ip) port 22.
Then I would like an easy way of using this port in the web interface.

So I added git.sshDisplayHost/Port to the config.

----
Commit comment:

Added git.sshDisplay{Port|Host} to hide port forward.

Running gitblit in a container it's easy to expose the ssh on the default port.
Using git.sshDisplayPort/git.sshDisplayHost you can expose the forwarded address
as the official location.

### DIFF
--- a/src/main/distrib/data/gitblit.properties
+++ b/src/main/distrib/data/gitblit.properties
@@ -110,6 +110,22 @@ git.sshPort = 29418
 # RESTART REQUIRED
 git.sshBindInterface = 
 
+# Specifies the ssh postname to publish in the web interface.
+# You may specify a hostname if you're forwarding from a different
+# hostname than web.canonicalUrl points to
+#
+# SINCE 1.6.3
+# RESTART REQUIRED
+git.sshDisplayHost = 
+
+# Specifies the ssh port to publish in the web interface.
+# You may specify a port that has been forwarded to git.sshPort
+# Recommended value: 22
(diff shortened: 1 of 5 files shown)


gitblit:
I understand the use-case for the port.  I am not understanding the hostname setting.  When would the hostname in `web.canonicalUrl` be different than the one in `git.sshDIsplayHost`?

bogeskov:
The reason for the git.sshDisplayHost, is that the primary url is mapped into an apache with proxypass together with some other services (/artifactory, /jenkins, /moinmoin , /sonarqube, ...), so that I have only one http entrypoint. But the ssh is on the server that actually runs the docker gitblit container.
I suppose I could install a port forwarder on port 22 on the webserver machine, and then add an extra ip, to that one for ssh access. However if I was to go that way, I would should probably rethink the entire setup, which hash all the symptoms of a but shooting organism.

bogeskov:
Just realized that I didn't handle empty settings properly. Committed a fix to that.

gitblit:
I manually merged & revised your contribution to the `develop` branch.  Thanks!

PR #269, opened 2015-06-09
Updated Dependencies

wangliyu:
Groovy 2.4.3
JGit 4.0.0.201506090130-r
SLF4j 1.7.12
Wicket 1.4.23

### DIFF
--- a/build.moxie
+++ b/build.moxie
@@ -103,11 +103,11 @@ repositories: central, eclipse-snapshots, eclipse, gitblit
 # Convenience properties for dependencies
 properties: {
   jetty.version  : 9.2.10.v20150310
-  slf4j.version  : 1.7.10
-  wicket.version : 1.4.21
+  slf4j.version  : 1.7.12
+  wicket.version : 1.4.23
   lucene.version : 4.10.0
-  jgit.version   : 3.7.0.201502260915-r
-  groovy.version : 2.4.1
+  jgit.version   : 4.0.0.201506090130-r
+  groovy.version : 2.4.3
   bouncycastle.version : 1.51
   selenium.version : 2.28.0
   wikitext.version : 1.4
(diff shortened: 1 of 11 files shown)


wangliyu:
As part of Eclipse mars release, the JGit is released 4.0, there is a big change:
https://dev.gitblit.com/commit/mirrors!jgit.git/686124bec32fcdee1545e7d9c312dcf0276b44af
 that cause gitblit compile error, also Groovy has updated to 2.4.3, Wicket 1.4.23, SLF4j 1.7.12, Please review and merge it in, also Since JGit 4.0, most of the class implemented AutoClosable interfaces, which means we can use try-with-resources feature from Java7. maybe need more tweak.

gitblit:
I don't know that I am ready for JGit 4.0 yet.  Maybe.

wangliyu:
I'll run the new change in my env for couple days see how it goes, I'm using GO version - with Java8.

wangliyu:
here is the highlight of changes in 4.0:
https://wiki.eclipse.org/JGit/New_and_Noteworthy/4.0

wangliyu:
I run the new build in my env, it works fine. clone/push/create branch/tag all are fine.

gitblit:
I want to sit on updating to JGit 4.0.  I know there is at least one intermittent [bug] in it that would affect my normal usage workflow.  As for Guice 4.0, you may or may not have noticed that Gitblit uses  a [patched build of guice-servlet](https://github.com/gitblit/google-guice).  In order to safely update to Guice 4.0 we'd need to apply that patch to the 4.0 code base.

The other dependency update proposals are fine.

cc @gzsombor

[bug]: https://bugs.eclipse.org/bugs/show_bug.cgi?id=422988

gitblit:
@wangliyu Please submit a new PR against `develop` without JGit 4 and without Guice 4.

PR #270, opened 2015-06-11
Upgrade jgit to 4.0

gzsombor:
Fixing the compilation failures, with using the AutoCloseable.close method and the 'try-with-resource' pattern.

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -41,14 +41,14 @@
 	<classpathentry kind="lib" path="ext/tracwiki-core-1.4.jar" sourcepath="ext/src/tracwiki-core-1.4.jar" />
 	<classpathentry kind="lib" path="ext/mediawiki-core-1.4.jar" sourcepath="ext/src/mediawiki-core-1.4.jar" />
 	<classpathentry kind="lib" path="ext/confluence-core-1.4.jar" sourcepath="ext/src/confluence-core-1.4.jar" />
-	<classpathentry kind="lib" path="ext/org.eclipse.jgit-3.7.0.201502260915-r.jar" sourcepath="ext/src/org.eclipse.jgit-3.7.0.201502260915-r.jar" />
-	<classpathentry kind="lib" path="ext/jsch-0.1.50.jar" sourcepath="ext/src/jsch-0.1.50.jar" />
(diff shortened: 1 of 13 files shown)


gitblit:
@gzsombor Thanks for taking the time to update to the `try()` syntax - it's definitely cleaner.  There were a couple instances where I thought the proposed changes introduced new potential failures due to relocation of code out of exception handlers.

I consider the update to JGit 4 fairly sensitive since I know it re-introduces an old JGit bug so of the three separate PRs proposing an update to JGit 4, I chose the one with the least number of diffs to make sure that I'm not introducing new potential failures into the code.

If you want to resubmit `try()` cleanup based on the current `develop`, that would be fine.  If you don't, I understand that too.

gzsombor:
Thanks, I think, I have the same problem (syncing between gerrit and gitblit sometimes fails with missing tree exceptions), hoped that upgrading jgit to 4.0 would fix it. Unfortunately not. And it's not clear for me, if it's a client side jgit error or a server side jgit error. I would be very happy if you could give some advice on it - yes, maybe it's not a good place for this kind of questions :)

gitblit:
Your missing tree sounds like [this JGit bug](https://code.google.com/p/gerrit/issues/detail?id=2296).  It's what I was referring to in #269.  Sadly JGit 4.0 purposefully restores this bug because the fix caused a performance regression observed in (large?) Gerrit installs.

The workaround for gerrit is to configure `receive.checkReferencedObjectsAreReachable=false`.  Gitblit 1.6 can configure this too, `git.checkReferencedObjectsAreReachable=false`.  I removed this setting in 1.7 after updating to the "fixed" version of JGit.  I have not yet reverted that commit on `develop`.

gzsombor:
Maybe I misunderstood something, but that flag is still used on the develop branch, here : https://github.com/gitblit/gitblit/blob/develop/src/main/java/com/gitblit/git/GitblitReceivePack.java#L131
And I can't find any other reference for that. I will try to disable this setting, and check if everything works as expected.

gitblit:
Ah.  Ok.  Maybe I just removed it from the documentation.  That would be good news.

PR #273, opened 2015-06-13
Refactor authentication for servlet HTTP header handler

mrjoel:
I'm working on integrating gitblit behind a reverse web proxy which performs authentication and provides group (team) membership via HTTP headers. As such, I'm trying to write a custom authentication plugin which pulls the user and team membership from the request headers. I wouldn't mind using the users.conf mechanism (ConfigUserService) for the storage of the gitblit specific authorization settings.

My goal is similar to current X509 handling, in that it will need to be verified on every request. I started by doing a proof of concept by hard-wiring it in AuthenticationManager, however I then followed it up with modularizing as in option C below, which I think is a better structured approach, because of the extensibility provided. Both steps are available as separate in sequence commits in this pull request.

The mechanism that I'm envisioning would be very similar to the following that we're using for XWiki:
        https://github.com/xwiki-contrib/xwiki-authenticator-headers

However, I was a bit unsure on the best location hooks to use to wire up a new auth plugin, where I have access to the HTTP request header values. After looking at the code base for a bit, I saw a few options:

        A. Write extension of UsernamePasswordAuthenticationProvider, but ignore the parameter values and pass back a UserModel based just on the request context headers.

I haven't found a way to access the request HTTP headers to perform the auth and verify information from within the AuthenticationProvider.authenticate(String username, char[] password) structure. Furthermore, it seems that this is only used for handling the login form submission, after which the session cookie is used for further requests. I'd like to avoid requiring a cookie at all, and build everything for a stateless session with a lifetime only for the duration of request processing.

        B. Modify existing AuthenticationManager direct handling of servlet authentication

The servlet authenticate() version seems to be called from SessionPage on every load, so that appeared to be a viable option. Wicket pages that inherit from SessionPage call GitblitManager, which in turn calls the servlet authenticate method on IAuthenticationManager. The direct handling of the servlet authenticate request within the AuthenticationManager could be updated to add the header authentication.

            Note that the javadoc for the IAuthenticationManager.authenticate() username/password method includes the following, but IUserService doesn't appear to have any authenticate methods (anymore?).
                * @see IUserService.authenticate(String, char[])

        C. Refactor AuthenticationManager so that the authenticate(HttpServletRequest httpRequest) flavor is pluggable like the username password mechanism. This would seem better than B for flexibility. Even adding the header auth into the existing flow in AuthenticationManager, it doesn't allow an installer to select the order or enabling/disabling of various mechanisms. For example, I want to configure it to only trust my custom headers managed by the proxy, not try X509 or pull values from the REMOTE_USER (as getUserPrincipal does).

It seems straight forward to add the servlet authenticate() overload to AuthenticationProvider, and implementations may provide a simple null returning implementation if they don't support a given authentication path. For now that would mean implementing it in my new plugin, and adding a simple null return to UsernamePasswordAuthenticationProvider.

The AuthenticationProvider provides user and team helper methods that check the checksum before calling the UserManager user/team update methods, but I'm just always blinding calling them directly from AuthenticationManager for every request (several per page load). That works but is quite inefficient on I/O, and making servlet authentication handling pluggable would enable using the helper methods as they currently stand within AuthenticationProvider.

This would also provide the benefit of each servlet mechanism being able to indicate their own support for support{DisplayName,Credential,EmailAddress,TeamMembership}Changes, whereas now it is required to be somewhat awkwardly handled in various permutations in the AuthenticationManager itself. The findProvider behavior of returning the NULL AuthenticationProvider if none other is found gets in the way here.

I'd appreciate any comments, there are a few items that I wasn't sure on, so I've submitted the pull request more as an RFC.
    
1. Should a specific AccountType be created for each provider method (what I've done), or should the existing EXTERNAL type by used?

2. The existing handling for AuthenticationManager uses flagSession which requires the AuthenticationType, so I added that to the AuthenticationProvider abstract, however it wasn't clear which values should be mapped through and how its handling differs from AccountType.

3. I haven't yet converted the existing servlet auth mechanisms to AuthenticationProvider yet, but I think for this to be complete that would need to be done as well. If you're okay with the approach I put together I will do that final conversion.

4. The existing requiresCertificate variant of the authenticate(httpRequest) method becomes a bit odd with the new construct. My thought is that this option ends up more cleanly supported as requiring only the (future reworked) X509AuthProvider loaded in order for the realm.authenticationProviders config value.
            
        For example, setting git.requiresClientCertificate previously would now look like:
                realm.authenticationProviders = x509client

        But optionally providing client cert, but not requiring it, looks something like:
                realm.authenticationProviders = x509client basicauth

        The 1.6.2 default processing order would look like:
                realm.authenticationProviders = container x509client sessioncookie basicauth

5. I'd also like to disable local auth entirely for my installation. The easiest way to do that seems to be move AuthenticationManager.authenticateLocal() away from an internal always executed call, and into an AuthenticationProvider module. Are there some assumptions baked in which I haven't noticed about auth providers being always external vs. local/internal authentication?

### DIFF
--- a/src/main/distrib/data/defaults.properties
+++ b/src/main/distrib/data/defaults.properties
@@ -817,6 +817,7 @@ realm.userService = ${baseFolder}/users.conf
 # Valid providers are:
 #
 #    htpasswd
+#    httpheader
 #    ldap
 #    pam
 #    redmine
@@ -1738,6 +1739,42 @@ realm.pam.serviceName = system-auth
 # BASEFOLDER
 # SINCE 1.3.2
 realm.htpasswd.userfile = ${baseFolder}/htpasswd
+
+# The name of the HTTP header containing the user name to trust as authenticated
+# default: none
+#
+# WARNING: only use this mechanism if your requests are coming from a trusted
+#          and secure source such as a self managed reverse proxy!
+#
+# RESTART REQUIRED
+# SINCE 1.7.0
+realm.httpheader.userheader =
+
+# The name of the HTTP header containing the team names of which the user is a member.
+# If this is defined, then only groups from the headers will be available, whereas
+# if this remains undefined, then local groups will be used.
+#
+# This setting requires that you have configured realm.httpheader.userheader.
+#
+# default: none
+#
+# RESTART REQUIRED
+# SINCE 1.7.0
+realm.httpheader.teamheader =
+
+# The regular expression pattern used to separate team names in the team header value
+# default: ,
+#
+# This setting requires that you have configured realm.httpheader.teamheader
+#
+# RESTART REQUIRED
+# SINCE 1.7.0
+realm.httpheader.teamseparator = ,
+
+# Auto-creates user accounts when successfully authenticated based on HTTP headers.
+#
+# SINCE 1.7.0
+realm.httpheader.autoCreateAccounts = false
 
 # Restrict the Salesforce user to members of this org.
 # default: 0 (i.e. do not check the Org ID)
--- a/src/main/java/com/gitblit/ConfigUserService.java
+++ b/src/main/java/com/gitblit/ConfigUserService.java
@@ -890,9 +890,6 @@ public class ConfigUserService implements IUserService {
 					user.displayName = config.getString(USER, username, DISPLAYNAME);
 					user.emailAddress = config.getString(USER, username, EMAILADDRESS);
 					user.accountType = AccountType.fromString(config.getString(USER, username, ACCOUNTTYPE));
-					if (Constants.EXTERNAL_ACCOUNT.equals(user.password) && user.accountType.isLocal()) {
-						user.accountType = AccountType.EXTERNAL;
-					}
 					user.disabled = config.getBoolean(USER, username, DISABLED, false);
 					user.organizationalUnit = config.getString(USER, username, ORGANIZATIONALUNIT);
 					user.organization = config.getString(USER, username, ORGANIZATION);
--- a/src/main/java/com/gitblit/Constants.java
+++ b/src/main/java/com/gitblit/Constants.java
@@ -568,7 +568,7 @@ public class Constants {
 	}
 
 	public static enum AuthenticationType {
-		PUBLIC_KEY, CREDENTIALS, COOKIE, CERTIFICATE, CONTAINER;
+		PUBLIC_KEY, CREDENTIALS, COOKIE, CERTIFICATE, CONTAINER, HTTPHEADER;
 
 		public boolean isStandard() {
 			return ordinal() <= COOKIE.ordinal();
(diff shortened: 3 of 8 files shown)


gitblit:
I'm ok with the approach you have coded so far to support HTTP header-sourced users.  I would prefer not to use the config settings of other realms in your new realm (e.g. `realm.container.autoCreate`).

In point 4 you have implicitly redefined an `AuthenticationProvider` to mean mechanism of authentication rather than provider of user authentication  (e.g. cookie, basic, certificate, etc VS ldap, htpasswd, PAM, Windows, etc).  If you want to make the supported mechanisms  configurable this will need to be a new set of classes/interfaces - like AuthenticationMechanism or something to that effect - and a different PR.

In point 5 you'd like to make `local` a separate authentication provider.  Conceptually I'm ok with that idea, but it will introduce even more permutations too an already complex system.  e.g. you can't manually add/create accounts without `local`.  You may (or may not) need `local` to administer teams.  We'd have to identify all the points `local` is implicitly relied on and handle them.  This would also need to be a separate PR.

As for your hope to remain stateless.... that'll be hard.  Wicket is a stateful framework and there are pages where this is the required behavior.  Gitblit does make an effort not to trigger all of Wicket's stateful record keeping behavior and generally succeeds, but you'll find that trying to avoid a `Session` is going to cause you pain.

mrjoel:
I've done some further work on the refactoring and turning everything into an AuthenticationProvider and ran into the issues you expected with mechanism vs. provider. I think it's still overall worth doing, but will take a bit more work. I'll break that off into a separate branch/PR, hopefully within a week or two. Meanwhile, I'll do some cleanup on the in-line HTTP header auth as is.

As for a separate realm autocreate flag, I had just used that one initially since it was there and easy. Creating a separate realm.X.autoCreate flag would seem a bit like proliferation of config options. One thing that I appreciate about the gitblit config file is that it's well documented, as easy to read through all of the options when doing an initial setup. Adding a bunch more options injudiciously would detract from this.

One may, however, legitimately want to auto create accounts for HTTP headers, LDAP, and PAM authentication, but not X509 authentication, so it does provide that flexibility. Perhaps a top-level value of realm.autoCreate which is used by each of the individual realms unless they are specifically defined. Let me know how you'd prefer this to be done.

On the statelessness, I think my biggest gripes aren't the state, but how wicket adds the JSESSIONID to the URL as well as cookie until the cookie is sent back. It ends up being ugly to map on the proxy side, so I'd like to just be able to disable that - but again, a separate PR.

mrjoel:
Also, what is the correct approach for AccountType for the HTTP header auth? The semantic difference between AccountType and AuthenticationType isn't clear.

The definition of AccountType in Constants.java contains EXTERNAL which would seem to include CONTAINER, LDAP, HTPASSWD and others, since they're all listed separately. However, the only use of AccountType.EXTERNAL is in ConfigUserService.java, with the following snippet:

`
if (Constants.EXTERNAL_ACCOUNT.equals(user.password) && user.accountType.isLocal()) {
    user.accountType = AccountType.EXTERNAL;
}
`

mrjoel:
Note that this PR is currently incompatible with https://github.com/gitblit/gitblit/pull/908, once that PR is addressed, I'll update this one to change flagSession to flagRequest.

mrjoel:
Rebased pull source branch against v1.7.1 tag (cleanly applies against master) which includes merge of #908

PR #274, opened 2015-06-13
Allow control over which internal repo URLs to show

mrjoel:
As asked on the list, I have gitblit running behind a proxy and have an external canonical path for repositories. I add them to otherUrls, but as things currently stand, there is no way to inhibit display of internal URLs if the required service is running.

I want to use the GitServlet, but only advertise the URL through an external endpoint.


gitblit:
Looks ok to me. Merged.

PR #275, opened 2015-06-29
use JEE 3.0 to allow cookie session tracking

mrjoel:
This is useful for us locally. It isn't supported for Tomcat 6, so depending on what installation requirements you want to have, may elect not to merge it yet while Tomcat 6 is still supported for security updates.


gitblit:
This is fine with me, thanks!

PR #277, opened 2015-06-29
NPE server fix

mrjoel:
When the shutdown port is already in use by something else, the port binding immediately above fails and generates an NPE.


gitblit:
:+1:

PR #281, opened 2015-06-30
allow internal privilege evaluation external URLs

mrjoel:
commit c20191fc0931a19bec0df1ab2b56f287e5d8b7c7 enabled support
for hiding internal URLs, but didn't consider that it broke the
evaluation of permissions (used for tickets, etc.), and caused
a NPE on repoUrl.permission when trying to view the TicketPage.

With all internal mechanisms disabled, it would result in the
first URL being external with unknown permissions. This adds an
option to use internal permissions even for external URLs.

Note that this does not grant any additional permissions, but
does offer the option to have gitblit advertise the full set of
what is allowed, even if the external URL imposes additional
restrictions.

### DIFF
--- a/src/main/distrib/data/defaults.properties
+++ b/src/main/distrib/data/defaults.properties
@@ -1203,6 +1203,20 @@ web.showGitDaemonUrls = true
 # SINCE 1.7.0
 web.showSshDaemonUrls = true
 
+# Should effective permissions be advertised for access paths defined in web.otherUrls?
+# If false, gitblit will indicate unknown permissions for the external link. If true,
+# gitblit will indicate permissions as defined within gitblit (including limiting to clone
+# permission is the transport type is not a valid push mechaism in git.acceptedPushTransports).
+#
+# Configure with caution: Note that gitblit has no way of knowing if further restrictions
+# are imposed by an external forwarding agent, so this may cause user confusion due to
+# more rights being advertised than are available through the URL. It will NOT grant
+# additional rights, but may incorrectly offer actions that are unavailable externally.
+# default: false
+#
+# SINCE 1.7.0
+web.advertiseAccessPermissionForOtherUrls = false
+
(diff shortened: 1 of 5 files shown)


src/main/java/com/gitblit/models/RepositoryUrl.java
```diff
@@ -41,8 +41,8 @@ public RepositoryUrl(String url, AccessPermission permission) {
 		this.permission = permission;
 	}
 
-	public boolean isExternal() {
-		return permission == null;
+	public boolean isPermissionKnown() {
```
gitblit: I would prefer `hasPermission()`

src/main/distrib/data/defaults.properties
```diff
@@ -1203,6 +1203,18 @@ web.showGitDaemonUrls = true
...
+# If enabled, a user will be offered actions for each "otherUrl" based
+# on the internally managed permissions.
+#
+# Configure with caution: Note that gitblit has no way of knowing if further
+# restrictions are imposed by an external forwarding agent, so this may cause
+# more rights to be advertised than are available. It will not grant additional
+# rights, but may incorrectly offer actions that are unavailable.
+#
+# SINCE 1.7.0
+web.useInternalPermissionsForOtherUrls = false
```
gitblit: Would this be a better name? `web.advertiseAccessPermissionForOtherUrls` and then describe that the configured push transports determine the advertised url permission?

src/main/java/com/gitblit/models/RepositoryUrl.java
mrjoel: Updated

src/main/distrib/data/defaults.properties
mrjoel: Yep, that reads more clearly. I also added the clarification on push transports.

PR #282, opened 2015-06-30
fix permission capping for HTTP/HTTPS

mrjoel:
Previously used request scheme, but request scheme is unrelated to
the URL being generated. Instead, base the permission capping on the
scheme of the URL itself.


gitblit:
Looks good.

PR #285, opened 2015-07-01
Emulate mount parameters for non-wicket capability

mrjoel:
Some of the one-off servlets are inconsistent when setting the web.mountParameters. This adds emulation support for having at least the project and repository name switch between query string/path. This eases implementing access control restrictions from an external reverse proxy.

### DIFF
--- a/src/main/distrib/data/defaults.properties
+++ b/src/main/distrib/data/defaults.properties
@@ -1159,7 +1159,11 @@ web.mountParameters = true
 
 # Some servlet containers (e.g. Tomcat >= 6.0.10) disallow '/' (%2F) encoding
 # in URLs as a security precaution for proxies.  This setting tells Gitblit
-# to preemptively replace '/' with '*' or '!' for url string parameters.
+# to preemptively replace '/' with a different character for url path parameters.
+# Note that '!' will always be accepted as input, however an additional character
+# may also be specified as accepted and to be used for generating links.
+#
+# Recommended characters are '!', ':', or '*'
 #
 # <https://issues.apache.org/jira/browse/WICKET-1303>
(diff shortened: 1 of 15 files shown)


mrjoel:
Note that HttpServletRequest.getPathInfo() is used for this functionality, however the current version of Guice as used by gitblit is buggy ( https://github.com/google/guice/pull/860, https://github.com/google/guice/issues/745).

The result is that any filename, path, or project that contains spaces or other URI encoded characters fail.

The fix from the pull request is included in the final 4.0 release of Guice. However, when I went to update the version being used, I saw that the version of com.google.inject.extensions:guice-servlet being used is noted as being a gitblit specific fork.

Is there a public git repo for the gitblit fork that describes the differences required? Rather than put in a workaround to gitblit, I'd rather just update the Guice dependency.

gitblit:
My fork is here: https://github.com/gitblit/google-guice

It's on my list to update to Guice 4 final.  Not sure when I'll get to it (help wanted).  My patch addresses google/guice#807 which is critical fix for enforcing Gitblit security.

Having said that, the scope of this PR makes me uncomfortable and for now I don't see this as a contender for merging.  Switching from query parameters to path parameters for repository names which may have subpaths and embedded forward-slashes needs to be carefully examined.  That has been the source of many problems over the years.

mrjoel:
Isn't the reason you have the web.forwardSlashCharacter to distinguish and separate parameters with embedded URLs from the request path? I followed that pattern assuming that the primary intent was to provide some level of disambiguation between nested levels of multiple subpaths for a repo especially.

gitblit:
It was added primary to deal with containers and proxy servers (Tomcat, Apache) that are openly hostile to the encoded forward-slashes used in mounted Wicket urls.

gitblit:
Thanks for making the effort but I'm going to pass on merging this.  Sorry.

PR #287, opened 2015-07-02
sort projects displayed on projects page

mrjoel:


gitblit:
AFAICT the list is already sorted.

mrjoel:
Definitely not sorted here in testing. I'm testing on Windows, so that could make a difference of native file directory ordering, or order created. I've also tested with git.cacheRepositoryList true and false to ensure that wasn't a factor.

PR #288, opened 2015-07-02
fix isPersonalRepository when user prefix is path

mrjoel:
The RepositoryModel uses the ModelUtils.isPersonalRepository by only passing
the project path instead of the entire project name as expected, so update
ModelUtils to handle this case.

### DIFF
--- a/src/main/java/com/gitblit/utils/ModelUtils.java
+++ b/src/main/java/com/gitblit/utils/ModelUtils.java
@@ -105,7 +105,7 @@ public class ModelUtils
 
 
 	/**
-	 * Exrtract a user's name from a personal repository path.
+	 * Extract a user's name from a personal repository path.
 	 *
 	 * @param path
 	 * 			A project name, a relative path to a repository.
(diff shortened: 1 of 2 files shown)


gitblit:
I don't understand this change.  In your test case `users` is not a personal repository IMO.

mrjoel:
True, the "users" itself is not a personal repository. However, there seems to be some inconsistency in what is expected to be used for that argument. Some places expect it to be a full repo path, others expect it to be some prefix of the path, which is what my fix addresses. The javadoc itself for the argument states that "A project name, a relative path to a repository. @return  true, if the name starts with the active user repository project prefix. False, otherwise.", implying that a partial path is enough to match. My change implements standard path semantics to normalize comparisons between paths that may differ only in trailing slash presence.

As mentioned in the request, RepositoryManager's isPersonalRepository just passes off to the ModelUtils version after empty check, but explicitly uses just the projectPath variable (which does not include any repository name itself, and does not include a trailing slash).

Likewise, I've found that my fix is incomplete, and ProjectManager.getProjectModel also uses just the project path as the argument to ModelUtils.getUsernameFromRepoPath. I've added a simple length comparison check to ModelUtils.getUsernameFromRepoPath, however based on the test case behavior it doesn't appear that this will ever match and return a username.

mrjoel:
Looking at this further, it also currently breaks when using subpaths for a user repo, but so does the getUserNameFromPath in that case too, so that likely warrants a separate cleanup PR (which may do the shifting to always pass a full path).

gitblit:
I've looked at this more closely.  You've identified a mismatch between how I had originally written user and project spaces and how they organically grew through contributions.  It's messy.  The original implementation consisted of the shipping default behavior: `~` connotes personal space, otherwise it's project space.

At present I think the proposed changes just mix up the mud.  I think the real fix is comprehensive overhaul of users vs projects - which I started in ticket 75 and never finished.

PR #290, opened 2015-07-09
only log successful ARF processing in debugmode

mrjoel:
be consistent with pattern in the rest of ARF to only log authentication results when running in debug mode


gitblit:
I don't know that I want to merge this.  Logging successful authentications gives admins an audit trail.

mrjoel:
I suppose that makes sense when that's the only logging mechanism, I was thinking of cases with hosted app and the container and/or proxy provided access log in front of the app itself. Perhaps a separate logger that can be turned down when running in such deployment scenarios?

gitblit:
Closing without merging for the aforementioned reason.

PR #296, opened 2015-08-10
Added better logging for fail2ban

Enrico204:
Hi,

I've modified the logging on SSH User-Password authentication to log IP addresses. It can be useful in software as fail2ban :-)


gitblit:
Thanks!

PR #915, opened 2015-09-10
Bump to Lucene 5.2.1

lucamilanesio:


gitblit:
This has been reverted because ticket queries no longer work.  This is solvable but it looks like it would require manually re-indexing all installations.  Of course an automated re-indexing trigger could be written, but at the moment I'm not interested in using my limited free time to write that.

lucamilanesio:
@gitblit yes, this is typically what we do with Gerrit: when people upgrade to a new version we generally ask to run a "reindex" to recreate the indexes. Do you believe this wouldn't be acceptable for GitBlit users?

gitblit:
I don't find it an acceptable requirement for Gitblit users.

Aside from the re-indexing, there appears to be some other necessary api/query changes required (see #918).  These are all minor, no doubt, but ATM I don't see any value in it for Gitblit aside from being able to say it's running the latest Lucene.

lucamilanesio:
@gitblit true, it isn't really a GitBlit requirement but rather needed for my plugin. Let me see if I can put a workaround instead in the GitBlit plugin build to avoid the clashes between different Lucene's versions.

lucamilanesio:
@gitblit I have analysed the situation: the original issue that drove me to the upgrade of Lucene is that ITicketService (despite the initial 'I') isn't an interface but an abstract class. Additionally it has a protected TicketIndexer that *IS* bound to the old Lucene's version.

I can workaround the problem by:
a) Rename ITicketService to AbstractTicketService and extract ITicketService as a pure interface
b) Bind the new pure ITicketService to something that does not link to the old Lucene's version when used inside Gerrit

In this way I'll obtain:
1) Neutral change in GitBlit, without having to upgrade Lucene
2) Ability to use the latest Lucene version when used as "Gitblit-plugin" for Gerrit

Would you agree on the approach? Shall I start a new PR with it?

Thank you in advance for your feedback.

Luca.

gitblit:
Yeah, that sounds like a much better plan.

PR #925, opened 2015-09-28
(ticket-537) user bindUser for group detection

RainerW:
Added new config key to enable disable the ldap user switch.


RainerW:
moved to new Branch for the github Ticket ID #247, because 537 was the google code ticket.

PR #934, opened 2015-10-01
Allow to strip domain from kerberos usernames

vitalif:
Add an option to strip domain from kerberos-authenticated usernames

### DIFF
--- a/src/main/java/com/gitblit/transport/ssh/SshDaemon.java
+++ b/src/main/java/com/gitblit/transport/ssh/SshDaemon.java
@@ -133,7 +133,7 @@ public class SshDaemon {
 		//Will do GSS ?
 		GSSAuthenticator gssAuthenticator = null;
 		if(settings.getBoolean(Keys.git.sshWithKrb5, false)) {
-			gssAuthenticator = new SshKrbAuthenticator(gitblit);
+			gssAuthenticator = new SshKrbAuthenticator(gitblit, settings.getBoolean(Keys.git.sshKrb5StripDomain, false));
 			String keytabString = settings.getString(Keys.git.sshKrb5Keytab,
 					"");
 			if(! keytabString.isEmpty()) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/gitblit/transport/ssh/SshDaemon.java
```diff
@@ -133,7 +133,7 @@ public SshDaemon(IGitblit gitblit, WorkQueue workQueue) {
 		//Will do GSS ?
 		GSSAuthenticator gssAuthenticator = null;
 		if(settings.getBoolean(Keys.git.sshWithKrb5, false)) {
-			gssAuthenticator = new SshKrbAuthenticator(gitblit);
+			gssAuthenticator = new SshKrbAuthenticator(gitblit, settings.getBoolean(Keys.git.sshKrb5StripDomain, false));
```
gitblit: Instead of passing a flag let's pass the IStoredSettings object and then retrieve the setting while inside the constructor.  This will be more flexible for the future.

vitalif:
Thanks for merging!:)

gitblit:
No problem.  I did back-off a change that was causing trouble for someone else.  If you are using krb5 I'd be interested to know if you see a behavioral difference when running the latest code.  I'm hoping it works the same for you.

vitalif:
What change do you speak of?

vitalif:
Do you mean 92ae83de6b4f5401a1007bbb26e2f01168e9d6cb?
As I understand it just means that gitblit push server will always use all authenticated methods including gssapi, publickey and password?

gitblit:
Yes, that is the commit.  I removed code from a contributor that was changing the default behavior of SSHD and breaking a collaborator - although I didn't have any problems with it.  I assume, since you opened a PR related to krb5, that you are using kerberos authentication.  If the `develop` branch after this commit is working for you, then all is well and I can put this issue to rest.

vitalif:
Just tried the newer version, everything is OK.

gitblit:
Bueno.  Thanks a bunch for confirming!

PR #935, opened 2015-10-01
Add "all repositories" parameter

vitalif:
Currently in Gitblit you must explicitly list all repositories to search in them. This does not allow to conveniently link to search page, for example, from a bugtracker. This patch adds "in all repositories" parameter so you can link to /lucene/?allrepos=1&query=$q or select a checkbox in the UI :)

### DIFF
--- a/src/main/java/com/gitblit/wicket/GitBlitWebApp.properties
+++ b/src/main/java/com/gitblit/wicket/GitBlitWebApp.properties
@@ -48,6 +48,7 @@ gb.password = password
 gb.tagger = tagger
 gb.moreHistory = more history...
 gb.difftocurrent = diff to current
+gb.allrepos = in all repositories
 gb.search = search
 gb.searchForAuthor = Search for commits authored by
 gb.searchForCommitter = Search for commits committed by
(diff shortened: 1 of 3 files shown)


src/main/java/com/gitblit/wicket/GitBlitWebApp.properties
```diff
@@ -48,6 +48,7 @@ gb.password = password
 gb.tagger = tagger
 gb.moreHistory = more history...
 gb.difftocurrent = diff to current
+gb.allrepos = in all repositories
```
gitblit: Let's tweak this to "All Repositories" so that it can be used elsewhere in the app, if needed.

gitblit:
@vitalif This looks great.  Make the small tweak and I'll merge.

vitalif:
There's an additional defect which should also be fixed: current implementation can duplicate the same repository multiple times in searchRepositories. In fact, you can even pass the same repository multiple times in the query parameter: /lucene/?query=xxx&repositories=repo+repo+repo, and it will lead to duplicate search results :)

PR #937, opened 2015-10-02
Revert "prevent session fixation for external authentication"

wangliyu:
with the change, it was cause issue when relogin with different user, to produce the issue, you can try logout current user and try login with different user, it will still remain the previous user after login success.


gitblit:
Hi @wangliyu I haven't tried Tomcat yet, but I don't see a problem with GO.

wangliyu:
Found the problem, in the users.conf, it remembered the session ID, so with the change, if the 2 users has same session cookie, it will always use the first one.

PR #943, opened 2015-10-16
Gitblit web site: Update Issues URL

StephenKing:
Gitblit web site: Update Issues URL


gitblit:
The site is built from Markdown sources in `src/site` and uses properties set in the `build.moxie` file.  In this case, I would take a look at the `build.moxie` file.

PR #945, opened 2015-10-17
Fix 2 possible NullPointer occurences

mawulf:


src/main/java/com/gitblit/utils/FileUtils.java
```diff
@@ -154,6 +155,14 @@ public static String readContent(File file, String lineEnding) {
 			System.err.println("Failed to read content of " + file.getAbsolutePath());
 			t.printStackTrace();
 		} finally {
+			if (reader != null){
```
gitblit: I don't believe closing the BufferedReader wrapper is necessary as the underlying stream is closed.

src/main/java/com/gitblit/manager/FederationManager.java
```diff
@@ -363,6 +363,10 @@ public boolean accept(File file) {
 							&& file.getName().toLowerCase().endsWith(Constants.PROPOSAL_EXT);
 				}
 			});
+			if (files == null) {
```
gitblit: This is a nice catch.

mawulf: Yes but this returns the empty initialized list before files(null) can throw a NPE. Or Not?

gitblit: Your proposed change is good.  Returning an empty list is what I want, not throwing an NPE.

PR #947, opened 2015-10-25
Log update for Fail2Ban usage

paulsputer:
Adds generic logging for failed authentication attempts across all authentication providers.  This provides a means to block IP addresses in the case of brute-force attacks on the main website.  Without this the website remains a potential attack vector straight through to internal authentication infrastructure such as LDAP.

Updated help page with additional regex and, building on #296, to use default GitBlit SSH Port 29418 to avoid any possible confusion.


gitblit:
Mucho Bueno.

Issue #949, opened 2015-11-08
Raw isen't working properly in gitlibt 1.6.2


paladox:
Raw isen't working here https://git.wikimedia.org/raw/mediawiki/extensions/DeletePagesForGood.git/master/DeletePagesForGood.hooks.php since Wikimedia updated gitblit to 1.6.2 from 1.4.

configs set like 

https://git.wikimedia.org/tree/operations%2Fpuppet.git/production/modules%2Fgitblit%2Ffiles

some files seems to work in raw https://git.wikimedia.org/raw/operations/puppet.git/production/modules!gitblit!files!gitblit.properties

gitblit:
Yes, I see.  There may be an issue resolving the difference between the "extensions" repo and the "extensions" subdirectory which DeletePagesForGood is part of.  Actually, I'm surprised Gitblit is still in service at Wikimedia.  I thought they jumped ship to Phabricator.

paladox:
Yes they are in the process on moving to phabricator but gitblit will be discontinued once gerrit migrated to phabricator on Wikimedia..

How would the raw problem be fixed. Would updating to the 1.7 snapshot work.

paladox:
Would it be because one of the configuations at https://git.wikimedia.org/tree/operations%2Fpuppet.git/production/modules%2Fgitblit%2Ffiles isen't updated or needs updated configs.

gitblit:
The 1.7 snapshot probably won't help here because of the repo name - directory name collision.  I think the problem you are observing here is there is a repo of the same name as a directory path (`mediawiki/extensions`).

When Gitblit parses `/raw/mediawiki/extensions/DeletePagesForGood.git/master/DeletePagesForGood.hooks.php` it will identify the repo by the first subpath match.

- is `mediawiki` a repo?  if not....
- is `mediawiki/extensions` a repo?  if not...
- is `mediawiki/extensions/DeletepagesForGood` a repo?  if not...

Once it has identified a repo it then identifies  the branch/ref followed by the path.  This could be fixed by writing urls using the forward-slash-character from the settings (e.g. `!` ) and properly parsing the repo from the url using this substitute - but that is not currently implemented.

I was not expecting users to create a repo and a subdirectory with the same name.  I believe the raw problem (and likely other problems too) could be fixed today in that installation by renaming the `mediawiki/extensions` repo to not collide with the `mediawiki/extensions/` directory.  Of course such a change creates downstream turbulence - git remote configs, etc.

[...]

PR #950, opened 2015-11-11
Match longest repository path towards the shortest

saper:
Try to to match longest repository paths first:

mediawiki/extensions/SemanticMediaWiki
mediawiki/extensions
mediawiki

Bug: https://github.com/gitblit/gitblit/issues/949

### DIFF
--- a/src/main/java/com/gitblit/servlet/RawServlet.java
+++ b/src/main/java/com/gitblit/servlet/RawServlet.java
@@ -161,23 +161,14 @@ public class RawServlet extends DaggerServlet {
 		}
 
 		// determine repository and resource from url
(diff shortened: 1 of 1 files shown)


saper:
I think the better fix would be to move all the logic to something like `RepositoryManager.getRepositoryFromURLPath(String)`, but I'm too unfamiliar with the code to propose a patch.

paladox:
+1 I haven't tested but looks good.

And gitblit really does need support for these paths.

gitblit:
@saper Thanks.  Looks good, works fine.

saper:
Thank you!

Issue #968, opened 2015-12-03
Git LFS push fails with server exception


paulsputer:
Git LFS failed to push and returned server error.  Gitblit logs show exception occurs:

```
2015-12-03 13:54:45 [WARN ] /git/r/qms/qms.git/info/lfs/objects/batch
java.lang.NullPointerException
        at com.gitblit.models.UserModel.canPush(UserModel.java:398)
        at com.gitblit.manager.FilestoreManager.addObject(FilestoreManager.java:152)
        at com.gitblit.manager.GitblitManager.addObject(GitblitManager.java:1258)
        at com.gitblit.servlet.FilestoreServlet.doPost(FilestoreServlet.java:124)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:707)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
        at com.google.inject.servlet.ServletDefinition.doServiceImpl(ServletDefinition.java:287)
        at com.google.inject.servlet.ServletDefinition.doService(ServletDefinition.java:277)
        at com.google.inject.servlet.ServletDefinition.service(ServletDefinition.java:182)
        at com.google.inject.servlet.ManagedServletPipeline.service(ManagedServletPipeline.java:91)
        at org.apache.wicket.protocol.http.WicketFilter.doFilter(WicketFilter.java:294)
        at com.gitblit.servlet.AccessRestrictionFilter.doFilter(AccessRestrictionFilter.java:253)
        at com.gitblit.servlet.EnforceAuthenticationFilter.doFilter(EnforceAuthenticationFilter.java:99)
        at com.gitblit.servlet.ProxyFilter$1.doFilter(ProxyFilter.java:89)
        at com.gitblit.servlet.ProxyFilter.doFilter(ProxyFilter.java:92)
        at com.google.inject.servlet.ManagedFilterPipeline.dispatch(ManagedFilterPipeline.java:120)
        at com.google.inject.servlet.GuiceFilter$1.call(GuiceFilter.java:133)
        at com.google.inject.servlet.GuiceFilter$1.call(GuiceFilter.java:130)
        at com.google.inject.servlet.GuiceFilter$Context.call(GuiceFilter.java:203)
        at com.google.inject.servlet.GuiceFilter.doFilter(GuiceFilter.java:130)
        at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1652)
[...]

PR #969, opened 2015-12-03
fix for #968 extracting repository from URI

paulsputer:
This fix allows the repository to be correctly identified by the `FilestoreServlet` when `server.contextPath = /git`


gitblit:
Nice catch.

PR #972, opened 2015-12-04
Disable HARDWRAPS markdown extension

miquella:
The HARDWRAPS extension is intended to be used like GitHub Flavored Markdown (GFM). GFM only uses hard wraps in comments, whereas the markdown rendered by GitBlit is generally documents not comments.

### DIFF
--- a/src/main/java/com/gitblit/utils/MarkdownUtils.java
+++ b/src/main/java/com/gitblit/utils/MarkdownUtils.java
@@ -17,6 +17,7 @@ package com.gitblit.utils;
 
 import static org.pegdown.Extensions.ALL;
 import static org.pegdown.Extensions.ANCHORLINKS;
+import static org.pegdown.Extensions.HARDWRAPS;
 import static org.pegdown.Extensions.SMARTYPANTS;
 
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


gitblit:
This disables wraps for everything. Maybe we need two methods instead of reversing the behavior for all uses.

miquella:
@gitblit: ah, good point.

I'm happy to extend this, what things are you thinking need to not change? My only intention was to change documentation type markdown rendering, but I'm not completely familiar with Gitblit, so I'm not sure what areas allow for comments and would need to stay the same.

gitblit:
I think the change here may be small - if we put it in the right place.  Like GH, GB allows for Markdown in Ticket comments.  Assuming your assertion about line wraps is correct - and I have to admit I've noticed some line-ending inconsistencies so I'm sure you are right - the `MarkdownUtils.transformGFM()` entry-point needs to preserve hardwraps.  This method which is used to Tickets eventually calls `MarkdownUtils.transformMarkdown(String, LinkRenderer)`, the one you proposed changing.

I'm flexible on an exact solution here.... new method, new arguments, whatever.

Issue #967, opened 2015-12-02
Git LFS Filestore menu item not shown for some users


paulsputer:
The Filestore menu item is not shown for users when part of a team membership.

This should be visible (perhaps optionally) to allow them to access the help section on using GitLFS.

gitblit:
I restricted access to this page.  In the 1.7 release it requires admin permissions.  The help content should probably be moved to a more accessible location like perhaps the site documentation.

paulsputer:
ahh thanks James just spotted the code the second I got the notification of your message ;)

Yea, how about I add the help page to the site documentation as you suggest, and perhaps update the menu item to show the help page directly of the commands to use - _similar to the way Gitblit provides helpful commands when working on the repo and ticket pages keeping the learning curve low for newer users_ - if it's a non-admin and then show the list if it's an admin?  

I could also add a config option to hide the button in the case where `authenticate view pages` is false yet we want to hide that page if needed.

I had thought it could also be useful at some point for users to see the list of LFS items in the case of difficulties uploading/downloading LFS items, but it seems like a lot of overhead for limited benefit at this stage with the filtering required.

gitblit:
>Yea, how about I add the help page to the site documentation as you suggest, and perhaps update the menu item to show the help page directly of the commands to use - similar to the way Gitblit provides helpful commands when working on the repo and ticket pages keeping the learning curve low for newer users - if it's a non-admin and then show the list if it's an admin?

Make it so.

> seems like a lot of overhead for limited benefit at this stage with the filtering required

From experience I know that if you advertise something a user may not have access to that triggers new forum threads and new bug reports. :stuck_out_tongue:   Better to be proactive & conservative here.

On the flip-side it may not be that expensive to filter the filestore objects list by access for presentation; that will depend on the number of blobs in the filestore.

paulsputer:
Thanks for merging in

PR #973, opened 2015-12-05
fix for #967 filestore menu and help

paulsputer:
Adds menu for non-admin users and relocates configuration help to site pages as discussed in #967.


gitblit:
If we filter blobs by view permissions then I'd re-enable the file store page for users. Till we do that this page will remain admin only.

paulsputer:
Ok will do.

paulsputer:
@gitblit  Turned out more straightforward than I was expecting :)

gitblit:
Looks good.  Nicely done!

Issue #498, opened 2015-08-12
com.gitblit.client.GitblitClient login operation swallows Exceptions


gitblit:
Originally reported on Google Code with ID 202

```
What steps will reproduce the problem?
1. Initialize an object of com.gitblit.client.GitblitClient
 and invoke login operation
2. Try to login with invalid credentials

What is the expected output? What do you see instead?
Client recieves no error message, but should return an exception. So seems UnauthorizedExceptions
and some other exceptions are swallowed there.

What version of the product are you using? On what operating system?
API 1.2.1  on Ubuntu 11.10

Please provide any additional information below.


```

Reported by `sajhak` on 2013-02-09 20:52:58

fzs:
Fixed by PR #979

PR #979, opened 2015-12-08
Remove empty catch blocks to stop swallowing exceptions. #498

pingunaut:
Removed empty catch blocks to enable the caller to handle possible exceptions. 
Unit tests run like before, gitblit manager also.

Regards and thanks for sharing gitblit,
Martin


gitblit:
Looks good, thanks.

PR #981, opened 2015-12-09
Log username with access logging

mrjoel:
Since we have the user available, log it with the message to make the access log more useful


gitblit:
A useful improvement.  Thanks.

Issue #978, opened 2015-12-07
Milestone due date selector with calendar view


paulsputer:
Milestone due date would be easier to use if it displayed calendar view.  Could use inbuiult HTML5 date selector.  According to http://caniuse.com/#search=date support is already available in Edge, Chrome, Opera, Safari and Android.  For those on IE and Mozilla it would automatically remain as text. @gitblit Happy for me to do a PR for this?

gitblit:
Yes, please.

paulsputer:
Merged in PR #982

PR #982, opened 2015-12-09
fix for #978 - HTML5 date input support

paulsputer:
+ JS patch/hack to coerce legacy wicket into talking to a HTML5 input type
+ JS script to hide inline help on date format when using HTML5 date picker
  + Date picker shown in user locale and standard does not support custom
  format.
  + Always sent in ISO8601 format

### DIFF
--- a/src/main/java/com/gitblit/wicket/Html5DateField.java
+++ b/src/main/java/com/gitblit/wicket/Html5DateField.java
@@ -1,0 +1,156 @@
+package com.gitblit.wicket;
+
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
+
+import org.apache.wicket.Session;
+import org.apache.wicket.markup.html.form.TextField;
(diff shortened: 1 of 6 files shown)


gitblit:
:+1: Gorgeous.  Now if only all the browsers had the Chrome data picker.

bviktor:
Does that work in Blink-based browsers, e.g. Opera?

gitblit:
Blink-based, should.  That's Chrome's engine now, right?  Presto-based, no idea; I no longer test Opera.

paulsputer:
@bviktor according to http://caniuse.com/#search=date Opera is supported, though I've not confirmed this.  I tested:

Chrome - Supported
Edge - Supported
Firefox - Fall back to textbox (see https://bugzilla.mozilla.org/show_bug.cgi?id=825294)
IE - Fall back to textbox

bviktor:
Opera is Blink-based :) Thanks for the report, @paulsputer , and also for all these nice additions recently :)

PR #983, opened 2015-12-10
revise logic for forcing dotfile to text

mrjoel:
Try to find a better happy medium for defining and offering "dotfiles" through the raw servlet. Change logic to only force text/plain when it is a classic dotfile, with leading dot and other other extensions.


gitblit:
:+1: Thanks for the fix!

PR #990, opened 2016-01-04
Fix authentication failure warning log messages for FEDERATION_USER

dbywalec:
The AuthenticationManager did not encounter for FEDERATION_USER and would unnecessarily
generate a lot of failure warning log messages, e.g:

Failed login attempt for $gitblit, invalid credentials from XXX.XX.XX.XX

A simple condition will prematurely return null bypassing the regular authentication path
and immediately make the authentication be routed via FederationManager.


gitblit:
Super!  Thank you.

PR #991, opened 2016-01-04
Lower log level of servlet authenticate when request is already authenticated

dbywalec:
When calling a servlet which has already been authenticated, the server would produce
a lot of superfluous log entries, e.g:

Called servlet authenticate when request is already authenticated.

The log level for this log entry has been lowered down to DEBUG.


gitblit:
Looks good, thanks!

Issue #993, opened 2016-01-05
rpc LIST_BRANCHES not possible without roles


stummb:
LIST_REPOSITORIES can be called by a user with #none roles, but LIST_BRANCHES not. In the documentation is stated that both require no special privileges.

We currently use version 1.6.2

gitblit:
Fixed by #994.

PR #994, opened 2016-01-06
Fix for #993 LIST_BRANCHES without admin

stummb:
Replaced the canAccess() method in RpcFilter with !adminRequest,
that should solve the problem.


gitblit:
Looks good.

PR #995, opened 2016-01-07
Add SQLite authentication realm

RedShift1:
These commit add an SQLiteAuthProvider. It uses Apache-style passwords so the same database can be re-used for authentication in Apache httpd.

It also splits off the actual validation of passwords to a new class, PasswordUtils. This was done because the password validation is the same as HtpasswdAuthProvider. Once this code has been tested, HtpasswdAuthProvider could leverage PasswordUtils to eliminate code duplication.

### DIFF
--- a/HOME.md
+++ b/HOME.md
@@ -33,6 +33,8 @@ This documentation is the source content from which the [Gitblit website](http:/
 [[src/site/setup_viewer.mkd]]
 [[src/site/administration.mkd]]
 [[src/site/setup_scaling.mkd]]
+[[src/site/setup_filestore.mkd]]
+
 
 ### Gitblit Tickets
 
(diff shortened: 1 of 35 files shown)


gitblit:
This could really be a JDBC auth provider and the user provides the driver. I'm a fan of SQLite but we'd be adding a 6MB dependency for very minimal gain. That is unless this can store all user and team data which at present it can not.

RedShift1:
I understand, I'll re-write this to generalize into a JDBC authprovider. I'm not sure how we'll have the user provide the necessary libraries during runtime but I'll figure something out.

gitblit:
Providing jars is easy in GO and generally easy in full containers.

Issue #998, opened 2016-01-08
Eliminate use of Weak DH Keys


kstange:
GitBlit GO 1.7.1, running on OpenJDK 1.8.0-65 is apparently still using weak 1024 bit keys from a common group.  Please consider updating the SSL parameters with 2048 bit keys or generate custom DH parameters for each installation.

RedShift1:
If I recall correctly, the upper limit for Java 7 DH group is 1024 bits. If hardcoded to 2048 bits, Gitblit would no longer run on Java 7 platforms.

gitblit:
@RedShift1 is correct.  It looks like J8 defaults to 1024 but you can set it to 2048 from the command-line. `-Djdk.tls.ephemeralDHKeySize=2048`.

RedShift1:
We can't fix this on Java 7, but we can document it. PR #1004

kstange:
Thanks for the solution.  Since I'm using Java 8, I've added the -D line to my startup script!

PR #1004, opened 2016-01-17
Ticket #998 document dh keysize

RedShift1:
Document how stronger DH keys can be generated when self-signed certificates are generated. Also fixes an unrelated typo.


gitblit:
Looks good.

Issue #1006, opened 2016-01-19
Spaces in URL


ManniManfred:
Hi, I have Problems cloning a Repository, that is in a folder with a space.
I tried 
a) git clone http://entwsrv01/r/Tests/Leerzeichen%20Test/Test1.git
b) git clone "http://entwsrv01/r/Tests/Leerzeichen Test/Test1.git"
both results in "fatal: repository ... not found"
What I'm doing wrong? Or isn't it supported?

gitblit:
Whilespace is illegal in repo names.

http://gitblit.com/administration.html#H3

PR #1007, opened 2016-01-19
Support whitespace in repository URL's

ManniManfred:
#1006


paulsputer:
Thanks for the PR, it's not quite as simple as that though and there's certainly some regex that would need updating.  Mainly though I think it will create a lot of new bugs and would allow users to break compatibility with connected systems, scripts, workflows etc.  Looking at GitHub, it converts spaces to hyphens, probably to avoid such issues.

Issue #974, opened 2015-12-06
Support in browser support for editing files


paladox:
Please support in browser editing of files and adding files please. Like gerrit has implanted this support gitblit should too. It would make everything easier.

gitblit:
Hi paladox. Thanks for the idea. Somewhere in the imported issues there is an older request for the same thing. I won't be working on this feature any time soon. If you'd like to submit a PR with a proposed implementation I'll be happy to review.

paulsputer:
:+1: This would be a useful feature to have.  It would be great to have something to quickly modify documents shown on the pages tab.  I had thought this may be considered differently for files and documents but I noticed a modern visual markdown editor [SimpleMDE](http://nextstepwebs.github.io/simplemde-markdown-editor/) uses codemirror behind the scenes so perhaps it's useful to consider both use-cases if selecting a client side library.

@gitblit would you be happy if I were to import [SimpleMDE](http://nextstepwebs.github.io/simplemde-markdown-editor/) into GitBlit?  If so I could do a PR for editing documents in the pages tab :)

paladox:
@paulsputer yes please could you import at least it is some support for gitblit if you were to do that.

The editor could be made more advanced over time.

Thanks.

gitblit:
@paulsputer Yup, that would be fine.  How about you work in my repo using pull requests rather than your fork.

paulsputer:
Great stuff, many thanks @gitblit will do :)

paladox:
Thanks.

paulsputer:
I tried to add  (https://github.com/eirslett/frontend-maven-plugin) as a maven dependency via moxie to it would download npm seamlessly but couldn't get it to work.  @gitblit I don't have much experience with moxie config so you may have more success with getting the plugin working.

Currently node/npm must be installed manually on the development system (https://nodejs.org/en/).  This handles the JS dependencies CodeMirror and SimpleMDE and ensures only one version of CodeMirror is used. Perhaps this configuration with npm could also provide a starting point for future Gitblit 2.x developments...

paladox:
@paulsputer thanks for working on this. Is it possible that a user does not have npm installed to use an alternative syntax highlighter code editor.

paulsputer:
[...]

PR #1022, opened 2016-02-24
974 in browser doc editor

paulsputer:
Document editor support for (#974) using ProseMirror to support wider range of use cases in the longer term compared to original PR (#1009) with SimpleMDE.  Markdown activated in this PR by default though other formats can be supported.  Code editor to be addressed in a later PR using CodeMirror. 

@gitblit I've added in merge/rebase capabilities with as minimal changes as possible. If merge fails it currently returns an error and the changes are lost which should be rare but could be rather frustrating.  What would you recommend for this?


paladox:
+1 thanks. I haven't tested but thanks for working on a built in editor.

paladox:
Does this support an option to disable the editor if gitblit acts as a mirror.

paulsputer:
At the moment the logic prevents a user from making a commit if they don't have write permissions.  So that should gracefully handle the mirroring case. It would probably be more intuitive if it also disables the links to the edit page and the edit page itself in this case.

paulsputer:
@gitblit would be great to get your feedback so we can merge in :)

gitblit:
Sorry Paul. I'm quite negligent and I'm traveling this week so I doubt I'll
get to it before next week. I trust your judgement so if you are ready to
merge and it's holding up your progress then go ahead and merge.

paulsputer:
No worries James, thanks.

paladox:
@paulsputer thanks.

Also do you have a screenshot of what it looks like please.

paulsputer:
Hi @paladox here we go, I'm sure they'll be improvements to the UI as it gets used more and other formats and features are added. :)

## Full screen editing mode
doceditortest-fullscreen

## Normal editing mode - with Heading 1 currently selected
doceditortest-activeh1

## Commit dialog with default commit message
doceditortest-commit

paladox:
@paulsputer yes probably but it already looks good, Thanks for working on it. It now allows users that are out about like on a mobile or tablet to be able to work on commits so thanks for allowing users to do this.

Issue #962, opened 2015-11-28
Ticket patchset revision delete functionality


paulsputer:
Adding functionality to delete patchset revision as discussed: https://groups.google.com/forum/#!topic/gitblit/JKh92WSQ_d4

Key aspects:
+ Update git revision reference using `update-ref` command
+ Update ticket journal
  + Should maintain append-only concept of journal
+ This interface would prevent deleting merged history, i.e can only delete rebased non-referenced items

PR #1039, opened 2016-04-03
Fix for #962 - Delete patchset ability

paulsputer:
Once a patchset has been deleted a manual `git gc` will need to be run on the server unless the JGit garbage collector is enabled.


paulsputer:
### Screenshot demonstrating patchset delete
deletepatchset

Issue #1037, opened 2016-04-01
Using tickets in private repositories


mliebelt:
We are using GitBlit 1.6.2, and have enabled tickets on one server (due to the demand of one project). We use the projects and repositories in a multi-tenant way, which means that only some groups of users are allowed to see the repositories.

I have tried now the following:

* Went to a private repository.
* Selected there "tickets"
* Created a new ticket.
* Opened a different browser.
* Logged in with a different user that has no access to the private repository before.
* Went to "my tickets"
* Searched there for something I knew was included in the ticket in the private repository.
* Found the new created ticket. (which is wrong in my opinion)
* When selecting the ticket, I get an error message: `Unauthorized access for repository <PROJECT/repo>.git` (which is ok)

In my opinion, tickets should behave in the same manner as the other artifacts in GitBlit. You should only have access to tickets, which are in repositories you could at least see (have view access).

paulsputer:
Hi @mliebelt thanks for reporting. I've just tested and experience the same issue.

I agree if the access policy is  `Restrict View, Clone, & Push` and the user has no view permission then this is an error.  Ticket results from lucene appear not to be filtered on any permissions.

This is also demonstrable for users who do not have permission to view non-private repositories.

gitblit:
Ouch.  Hard to believe I let that slip by.  Thanks @mliebelt for reporting and @paulsputer for fixing!

paulsputer:
Fixed in PR #1040

mliebelt:
Thank's a lot for the answers, and of course for the fix. One more reason to upgrade earlier (when it is contained in a new version).

Great tool, with great support!

PR #1040, opened 2016-04-04
Fix for #1037 myTickets now honours permissions

paulsputer:
Ticket search results are now only displayed if user has view rights to the repository.

This also corrects the handling of tickets that were created by, watched by, assigned to or mentions a user that no longer has view rights.


paulsputer:
@gitblit we may want to consider a minor release for this change as it fixes a potential information leak.  What do you think?

gitblit:
I can spin up a release whenever you think it's needed, but based on what has merged into master I would probably bump it to 1.8.0.

paulsputer:
Ok thanks, in that case I'll get a few small changes sorted i'll list those in a 1.8.0 prep ticket

Issue #1042, opened 2016-04-06
Intrusive  Filestore lozenge


paulsputer:
When viewing a tree or commit page with a large number of filestore items the lozenge indicating that the item is in the filestore appears intrusive.  This should be changed to something less smaller and less such as an icon with tooltip so as to provide more screen real estate to other items.

PR #1044, opened 2016-04-06
Fix for #1042 - Filestore items now shown as icons

paulsputer:


paulsputer:
### Before
before

### After
after

PR #1046, opened 2016-04-07
Fixes #1028 - FilestorePage now pages and filters

### LINKED ISSUE #1028: Filestore page - Items not paging
As more files are being added to the filestore all objects are listed on a single page with no paging mechanism.  The list view is useful when first using the filestore but quickly becomes difficult to yield useful information.

Perhaps a more useful would be to display stats similar to the repository metric page to give a feel for utilization and identify only those objects in a uploading or error status.

paulsputer:


paulsputer:
### Before
filestorebefore

### After

filestoreafter

gitblit:
Nice!

PR #1047, opened 2016-04-08
Fix build of javadoc in Windows(zh-cn)

yyjdelete:
Specify encoding of javadoc to "utf-8".
That will fix build of javadoc in Windows zh-cn (and maybe some other locale).
Unknown why javac works well.

Tested with oracle jdk1.8.0_77

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -433,7 +433,7 @@
 		</zip>
 		
 		<!-- Build API JavaDoc jar -->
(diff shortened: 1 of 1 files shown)


yyjdelete:
Found encoding of javac is set at
https://github.com/gitblit/gitblit/blob/master/build.moxie#L77

paulsputer:
Thanks

Issue #1048, opened 2016-04-08
Option to reference ticket without closing


paulsputer:
When working within the branching workflow of tickets, commits are shown on the ticket activity page and commit listings link to the ticket directly via a lozenge tag.

When working outside the branching workflow (i.e commiting hotfixes or blobs) referencing a ticket does not update ticket activity and no direct link is added to the ticket when listing the commits.  The only link is from the commit message to the ticket, and therefore difficult when auditing traceability of changes.

paulsputer:
All appears good after testing and usage

paulsputer:
All appears good after testing

PR #1055, opened 2016-04-11
Ticket Reference handling #1048

paulsputer:
+ Supports referencing:
  + from other tickets via comments
  + from commits on other branches

### DIFF
--- a/src/main/distrib/data/defaults.properties
+++ b/src/main/distrib/data/defaults.properties
@@ -574,6 +574,13 @@ tickets.requireApproval = false
 # SINCE 1.5.0
 tickets.closeOnPushCommitMessageRegex = (?:fixes|closes)[\\s-]+#?(\\d+)
 
+# The case-insensitive regular expression used to identify and link tickets on
+# push to the commits based on commit message.  In the case of a patchset 
+# self references are ignored 
+#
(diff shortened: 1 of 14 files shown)


paulsputer:
### Example showing reference by commit `8E5DA2` and `ticket/2`
ticketreferences

paulsputer:
@gitblit could you cast an eye over this and share your thoughts?

gitblit:
I will cast my eye.  Trying to make time.

gitblit:
Thoughts...

1. We need to unit test the new features of TicketModel and ITicketService.
2. I think the use of lozenges for commit links and ticket links overloads their appearance value.  I would advocate it would be better to use a plain, fixed-width font link for commit shas and a plain link for the ticket reference.

paulsputer:
Thanks James, (1) will do, (2) very true, I tried to use the lowest level lozenge but still looks a bit too much, I'll make those fixes and post back.

paulsputer:
ok @gitblit can you take another look at this and see if there's anything else we need to address on referencing in this PR? It's getting a bit larger than I was anticipating, then again half of it is testing the different use cases.

paulsputer:
Here's the updated UI
gitblitrefui2

PR #1087, opened 2016-06-14
Fixes #964 - add missing redirect after restoring user in new session

### LINKED ISSUE #964: Refreshing page after some time returns start page (dashboard)
Since the upgrade to version 1.7.1 i've noticed a little 'problem' when refreshing a page after some time. I have a project summary page open almost the whole day and if I leave it alone some time (i haven't measured the exact time, but 30 minutes max), come back and  refresh the page it returns the dashboard page. If I refresh a second time I get the summary page again.
I'm using Chrome and have gitblit running behind nginx, if this matters.

mereth:
Hi,

I may have a fix for #964.

If `web.authenticateViewPages` is set to `true`, and when the session is invalidated due to server restart or if the max inactive interval has been reached (default 30mn) :
* `AuthorizationStrategy.isPageAuthorized` save the target url and return `false` for any page except the homepage
* `AuthorizationStrategy.onUnauthorizedInstantiation` redirect _internally_ to the homepage
* `SessionPage.login` restore user informations and the homepage is rendered

It seems it was missing the redirect to the saved target url to render the correct page

Regards

### DIFF
--- a/src/main/java/com/gitblit/wicket/pages/SessionPage.java
+++ b/src/main/java/com/gitblit/wicket/pages/SessionPage.java
@@ -98,6 +98,7 @@ public abstract class SessionPage extends WebPage {
 					}
 				}
 				session.setUser(user);
+				session.continueRequest();
 				return;
 			}
 		}


paulsputer:
Thanks @mereth that's a great one-liner and does appear to have fixed #964

PR #1100, opened 2016-07-11
fix user effective permission display

felazuris:
There was issue if users is in multiple groups, then in access permissions tab under effective was shown only alphabetically first groups permissions, not the greatest ones. Fixed issue by adding some checks if permissions should be replaced in list. Maybe there is better way to fix this issue. (I am not really developer, just SCM engineer)


flaix:
The weird thing to me is that team permissions override specified user permissions. I believe it should be the other way around. But yeah, at least the displayed permissions should reflect the ones applied when checking access.

flaix:
Rebased and merged. Closing this as done.

PR #1105, opened 2016-07-29
Fix &rt; to &gt; in mail (in rename status line)

### LINKED ISSUE #1104: "-&rt;" instead of "->" in HTML mail after pushing a rename commit
In the mail (HTML) that Gitblit sends me automatically after a commit push, for a file renaming I see a line like this:

> original/path **-&rt;** new/path

which should rather look like this:

> original/path **->** new/path

guilliamxavier:
fixes #1104

### DIFF
--- a/src/main/distrib/data/groovy/sendmail-html.groovy
+++ b/src/main/distrib/data/groovy/sendmail-html.groovy
@@ -336,7 +336,7 @@ class HtmlMailWriter {
         }
         builder.td() {
             mkp.yield header.oldPath
-			mkp.yieldUnescaped "<b> -&rt; </b>"
+			mkp.yieldUnescaped "<b> -&gt; </b>"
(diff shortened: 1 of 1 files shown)


guilliamxavier:
Indeed the HTML entity for `>` is `&gt;`, not "`&rt;`" which is not an existing HTML entity _(for the record: the angle brackets `<` and `>` entities are `&lt;` and `&gt;` which stand for "**l**ess **t**han" and "**g**reater **t**han", not the hypothetical "`&lt;` and `&rt;`" that I suppose one could type thinking to "**l**ef**t**" and "**r**igh**t**" since e.g. `«` and `»` are `&laquo;` and `&raquo;` which stand for "**l**eft **a**ngle **quo**te" and "**r**ight **a**ngle **quo**te"...)_.

gitblit:
Thanks for the fix.

PR #1110, opened 2016-08-04
1.8.x tranditional chinese translation

WilliamFromTW:
1.8.x tranditional chinese translation after user experience


gitblit:
Thanks!

Issue #1137, opened 2016-10-17
Disable client certificates entirely


oddeirik:
I've recently configured Gitblit 1.8.0 with a proper, CA-signed SSL certificate. (This in itself was a bit of a pain trying to convert the certificate to a format that keytool and/or portecle could handle that _also_ included the private key, but #1050 got me on the right track).

So now all's well, right? Well, sort of. It turns out that some users are now asked which client certificate to use when opening the web interface. We're using Skype for Business via Office 365 and it seems that the Skype client installs a self-signed client certificate on the user's computer). Chrome and IE/Edge will pop up asking whether to use this certificate or not (while Firefox/Firefox Dev will not, curiously enough) when navigating to the Gitblit web interface. This is, of course, bound to cause problems because clicking Ok/Accept on this certificate will cause the page to fail to load (net::ERR_CONNECTION_CLOSED in Chrome). Closing the dialog or hitting Cancel works, but I'd rather not have that confusing dialog there at all :)

Looking through defaults.properties, I found this:

```
# If serving over https (recommended) you might consider requiring clients to
# authenticate with ssl certificates.  If enabled, only https clients with the
# a valid client certificate will be able to access Gitblit.
#
# If disabled, client certificate authentication is optional and will be tried
# first before falling-back to form authentication or basic authentication.
#
# Requiring client certificates to access any of Gitblit may be too extreme,
# consider this carefully.
#
# SINCE 1.2.0
# RESTART REQUIRED
server.requireClientCertificates = false
```

Is there a way to entirely disable client certificates entirely and not make them optional? And why is this suddenly happening because I've used a different SSL certificate for Gitblit?

`server.requireClientCertificates` is of course not overridden in my own configuration gitblit.properties file.

I don't think there's an easy way to disable the client certificates in Skype for Business either, not as long as it's hosted off of Office 365 at least.

gitblit:
No, not without hacking the [server launcher](https://github.com/gitblit/gitblit/blob/master/src/main/java/com/gitblit/GitBlitServer.java#L288-L292).  I have no idea why the Skype for Business cert confuses things.

oddeirik:
I'm not entirely sure why Lync/Skype has to install a client certificate either, but it's not trivial to disable for us it seems.

Would it be possible to add another setting called _server.wantClientCertificates_, and do something like this in GitBlitServer.java?

```
if (params.requireClientCertificates) {
	factory.setNeedClientAuth(true);
} else {
[...]

PR #1138, opened 2016-10-17
Option to explicitly enable or disable optional client certificates

oddeirik:
Ref #1137 this setting allows you to disable optional client certificate authentification as well.

I've tested this on my server and there's no longer any prompt to pick a user certificate when connecting to the web interface.

You might want to change the wording or the name of the setting to something you see more fit, I just tried matching the _requireClientCertificates_ setting.

### DIFF
--- a/src/main/distrib/data/defaults.properties
+++ b/src/main/distrib/data/defaults.properties
@@ -2059,8 +2059,8 @@ server.storePassword = gitblit
 # authenticate with ssl certificates.  If enabled, only https clients with the
 # a valid client certificate will be able to access Gitblit.
 #
-# If disabled, client certificate authentication is optional and will be tried
-# first before falling-back to form authentication or basic authentication.
+# If disabled, optional client certificate authentication is configurable by
+# server.wantClientCertificates
 #
 # Requiring client certificates to access any of Gitblit may be too extreme,
 # consider this carefully.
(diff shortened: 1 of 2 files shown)


mobyplum:
Did this PR make it in to 1.9.0?

fzs:
No, this is still open and on the list.

flaix:
I would prefer to not add a second option and expose the ambiguity of Java SSL in this case. I could imagine that the existing option is converted from a boolean to hold either an additional `off` value, or three completely different values like `required`, `optional`, `none`, or like Tomcat with `true`, `want`, `false`.
I guess the question is if this needs to be done in a backwards compatible way or if the meaning for `false` could be changed because everyone who has it set to `false` wants it `off` anyways.

KOTRET:
Converting to true / want / false would be a good solution, for me it seems to be compatible: only those who want to have this optional need to change that setting, which - I would think - is not the vast majority.

flaix:
Well, I had opted for the first one now, in the spirit of keeping backward compatibility. The three new values are `required`, `optional` and `none`. I did this so that an update would not change existing behaviour, which would be a surprise I myself don't really like. So the existing `true` and `false` values keep their meaning and are mapped like so:
`true` = `required`
`false`= `optional`

While it could be true that in reality hardly anyone would be affected, I unfortunately have no idea how many people still use Gitblit, i.e. how large the existing user base is that could be negatively affected. I don't think download numbers really mirror installed base.

Issue #1076, opened 2016-05-16
Slow startup of web / http manager 


ieugen:
Hello, 

I have configured gitblit on my machine and all was well until now. It seems whenever I start /restart the service it takes **very long** time to start SSH Manger and the web gui. 

As you can see, it takes `Started @136214ms` to start =136 seconds  

````
May 16 19:32:39 netdava.com java[2863]: 2016-05-16 19:32:39 [INFO ] ----[com.gitblit.manager.IServicesManager]----
May 16 19:32:39 netdava.com java[2863]: 2016-05-16 19:32:39 [INFO ] Federation passphrase is blank! This server can not be PULLED from.
May 16 19:32:39 netdava.com java[2863]: 2016-05-16 19:32:39 [INFO ] Fanout PubSub service is disabled.
May 16 19:32:39 netdava.com java[2863]: 2016-05-16 19:32:39 [INFO ] Git Daemon is listening on 0.0.0.0:9418
^[^[^[





May 16 19:34:51 netdava.com java[2863]: 2016-05-16 19:34:51 [INFO ] SSH Daemon (NIO2) is listening on 0.0.0.0:29418
May 16 19:34:51 netdava.com java[2863]: 2016-05-16 19:34:51 [INFO ]
May 16 19:34:51 netdava.com java[2863]: 2016-05-16 19:34:51 [INFO ] ----[com.gitblit.manager.IFilestoreManager]----
May 16 19:34:51 netdava.com java[2863]: 2016-05-16 19:34:51 [INFO ] No filestore metadata file found
May 16 19:34:51 netdava.com java[2863]: 2016-05-16 19:34:51 [INFO ]
May 16 19:34:51 netdava.com java[2863]: 2016-05-16 19:34:51 [INFO ] ----[com.gitblit.manager.IPluginManager]----
May 16 19:34:51 netdava.com java[2863]: 2016-05-16 19:34:51 [INFO ] PF4J version 1.7.1 in 'deployment' mode
May 16 19:34:51 netdava.com java[2863]: 2016-05-16 19:34:51 [INFO ] Enabled plugins: []
May 16 19:34:51 netdava.com java[2863]: 2016-05-16 19:34:51 [INFO ] Disabled plugins: []
May 16 19:34:51 netdava.com java[2863]: 2016-05-16 19:34:51 [INFO ] No plugins
May 16 19:34:51 netdava.com java[2863]: 2016-05-16 19:34:51 [INFO ]
[...]

PR #1140, opened 2016-10-27
Issue #1076: load commit cache in a background thread

tomaswolf:
* Make the CommitCache fully thread-safe. It was using a
  ConcurrentHashMap containing lists, but then handed out these lists.
  It also did multiple operations on that map that as a whole should
  be atomic.
  
* Use isEmpty() instead of size() == 0.

* Run the loading of the commit cache in a background daemon thread


gitblit:
:+1: Looks great!

PR #1141, opened 2016-10-27
Issue #1011: Resolve StackOverflowErrors on page serialization

tomaswolf:
They are caused by Gitblit pages holding references to JGit RevCommits. 

JGit commit objects are a recursive data structure; they have links to
their parent commits. Serializing a JGit commit will try to recursively
serialize all reachable ancestors as faras they have been loaded. If
that ancestor chain is too long, a StackOverflowError is thrown during
Wicket's page serialization if a page has a reference to sucha JGit
commit.

Fixed by making sure that pages o not contain references to JGit
commits. Use the (existing) wrapper object RepositoryCommit instead.
- RepositoryCommit has a transient reference to the JGit commit and
  reads the commit from the repository upon de-serialization.
- RefModel is a similar case (JGit tags/branches may also have links
  to the commits they point to). Solved a bit differently by making it
  a pure data object by transferring the interesting data from the JGit
  object in the constructor.
- Change DataViews instantiated with RevCommit to use RepositoryCommit
  instead.
- Change inner anonymous DataViews to ensure they do not have a
  synthesized field referencing the "allRefs" map. Such a synthesized
  field would also get serialized, and then serialize JGit commits
  again.

Finally, remove non-transient logger instances in Wicket classes. Those
might lead to NotSerializableException.

These StackOverflowErrors have been reported in several places since
2014:
- https://groups.google.com/forum/#!topic/gitblit/GH1d8WSlR6Q
- https://bugs.chromium.org/p/gerrit/issues/detail?id=3316
- https://groups.google.com/d/msg/repo-discuss/Kcl0JIGNiGk/0DjH4mO8hA8J
- https://groups.google.com/d/msg/repo-discuss/0_P6A3fjTec/2kcpVPIUAQAJ
- https://github.com/gitblit/gitblit/issues/1011
- https://github.com/tomaswolf/gerrit-gitblit-plugin/issues/21

### DIFF
--- a/src/main/java/com/gitblit/models/RefModel.java
+++ b/src/main/java/com/gitblit/models/RefModel.java
@@ -59,9 +59,9 @@ public class RefModel implements Serializable, Comparable<RefModel> {
 		this.reference = ref;
 		this.displayName = displayName;
 		this.date = internalGetDate(refObject);
-		this.name = (ref != null) ? ref.getName() : displayName;
+		this.name = ref != null ? ref.getName() : displayName;
 		this.type = internalGetReferencedObjectType(refObject);
-		this.objectId = internalGetObjectId(reference);
+		this.objectId = ref != null ? ref.getObjectId() : ObjectId.zeroId();
 		this.id = this.objectId.getName();
 		this.referencedObjectId = internalGetReferencedObjectId(refObject);
 		this.referencedId = this.referencedObjectId.getName();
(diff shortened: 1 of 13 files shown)


tomaswolf:
More details and analysis can be found at https://github.com/tomaswolf/gerrit-gitblit-plugin/issues/21#issuecomment-255589350 and following comments.

PR #1144, opened 2016-10-28
Improve logging when sending email fails.

fzs:
This PR adds error logging when sending emails from Gitblit failed because the SMTP server rejected the recipient. This makes it easier to find out what is going wring instead of just finding a stack trace in the log.
This patch is in relation to [ticket 224](https://dev.gitblit.com/tickets/gitblit.git/224). It does not add any throttling yet, just analysis.


gitblit:
:+1:  Good idea.

PR #1147, opened 2016-10-31
Fix disabled links in PagerPanel

tomaswolf:
Disabled links in the PagerPanel (used on the LuceneSearchPage to page
through search results) were only rendered as "disabled". The links
themselves remained active, which gives strange effects when clicked.
For instance it was possible to move to result pages -1, -2, and so on.

Really disable the links. Add missing CSS rules to have correct styling
as Wicket renders disabled links as spans, not anchors. Include the new
CSS file in BasePage.html. And add the left/right arrows only if not on
the first/last page.


gitblit:
Thanks for the fix.

PR #1152, opened 2016-11-18
Set "can admin" permission on LDAP users and teams correctly

fzs:
The canAdmin permission is set on a LDAP user, when the user is listed
in `realm.ldap.admins` or is a member of a team listed in `realm.ldap.admins`.
This leads to inconsistent and surprising behaviour on the EditUser page
when clicking the "can admin" checkbox. Also, the "can admin" checkbox
is disabled, but not checked, for teams that are listed as admin teams.

The new behaviour implemented in this patch makes users and teams from
LDAP match local ones. That means:
* LDAP teams that are listed in `realm.ldap.admins` get the canAdmin
  property set if teams are maintained in LDAP.
* LDAP users that are listed in `realm.ldap.admins` get the canAdmin
  property set if teams are maintained in LDAP.
* LDAP users do not get the canAdmin property set, if they are only a
  member of a team listed in `realm.ldap.admins`.
* The `supportsRoleChanges` method for users and teams of the
  `LdapAuthProvider` unconditially returns false if teams are
  maintained in LDAP, not only for users and teams listed in
  `realm.ldap.admins`.
* Therefore, for all LDAP users and teams the "can admin" checkbox
  is always disabled if teams are maintained in LDAP.

### DIFF
--- a/src/main/java/com/gitblit/auth/LdapAuthProvider.java
+++ b/src/main/java/com/gitblit/auth/LdapAuthProvider.java
@@ -171,6 +171,8 @@ public class LdapAuthProvider extends UsernamePasswordAuthenticationProvider {
 							final Map<String, TeamModel> userTeams = new HashMap<String, TeamModel>();
 							for (UserModel user : ldapUsers.values()) {
 								for (TeamModel userTeam : user.teams) {
+									// Is this an administrative team?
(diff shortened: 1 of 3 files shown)


gitblit:
Nice. :+1:

Issue #1156, opened 2016-11-29
Ldap groups query and Admin groups


jvaleriano:
I configure LDAP with correctly. 
I logIn with my user of AD (with Adminstration group) ands work ok, but wen I going to create a new repository and asigned to a group of AD I only can see MY GROPUS OF AD. The query to show ldap groups is the same that is used to check if you are administrator (realm.ldap.groupMemberPattern) and if I change it to filter the groups that start with GIT_* any logger user became an administrator (because the admins group is GIT_ADMIN). I try diferent way to config that but really the dropbox to select a goup for a repository must query the LDAP with another filter to separate the authentication for the 
![git_](https://cloud.githubusercontent.com/assets/11890664/20723342/e4f0ab6a-b648-11e6-8a82-45353638394f.png)
administration.

fzs:
Can you provide your configuration of `realm.ldap.groupMemberPattern`  and `realm.ldap.groupBase`?
The `realm.ldap.groupMemberPattern` is used to search for all teams that a certain user is a member of.
The teams and users with administrative rights are determined by full group or user names. Did you maybe not use the user name or DN in your `realm.ldap.groupMemberPattern` so that a user is always a member of all groups?

jvaleriano:
This is my config
realm.ldap.groupBase = OU=Grupos,OU=Argentina,DC=sis,DC=ad
realm.ldap.groupMemberPattern = (&(objectClass=group)(member=${dn}))
realm.ldap.admins = @GIT_ADMIN

Maybe i dont express well. The issue is if I'm administrator I want to see all groups of AD not only my groups. 
When I try to sort out the problem and see all groupstaht start with GIT_* in the selectbox.. I see that all user became an admin because realm.ldap.admins is GIT_*.... I can change that but the selectbox should show all groups not only where the admin is member.

jvaleriano:
// EditRepositoryPage
		
final RegistrantPermissionsPanel teamsPalette = new RegistrantPermissionsPanel("teams",
				RegistrantType.TEAM, app().users().getAllTeamNames(), repositoryTeams, getAccessPermissions());

here it get the class to load on the selectbox 

	/**
	 * Returns the list of all teams available _**to the login service.**_
	 *
	 * @return list of all teams
	 * @since 0.8.0
	 */
	List<String> getAllTeamNames();

fzs:
[...]

PR #1157, opened 2016-12-01
#1156

jvaleriano:
Is a minnor change but allow the Administrator see all groups that match with realm.ldap.groupAdminMemberPattern property.

### DIFF
--- a/src/main/distrib/data/defaults.properties
+++ b/src/main/distrib/data/defaults.properties
@@ -1899,6 +1899,23 @@ realm.ldap.groupBase = OU=Groups,OU=UserControl,OU=MyOrganization,DC=MyDomain
 # SINCE 1.0.0
 realm.ldap.groupMemberPattern = (&(objectClass=group)(member=${dn}))
 
+# Filter criteria for LDAP groups for Admins
+#
+# Query pattern to use when searching for a team. This may be any valid 
+# LDAP query expression, including the standard (&) and (|) operators.
+#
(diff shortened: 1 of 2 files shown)


fzs:
This patch is hard to read because it includes lots of lines where the only change seems to be whitespace and indentation change.
From what I can see it looks like your patch will add Gitblit administrators to every LDAP group. Which would solve your problem in #1156, but it doesn't fix the root of your problem, I believe. Plus, it is not a good idea, because team membership should be defined in LDAP.

jvaleriano:
My bad sorry for the identation.
the idea is made a cache only for the groups that Im going to use for the git. so I cant put the realm.ldap.synchronize in true. But I need that an administrator could use all the groups that match with a pattern like "git_*"

fzs:
The PR, as it is, cannot be merged. It adds all admins to all groups. Plus, by default, it synchronizes all groups from LDAP, even if synchronization is off. It tries to tackle the problem of seeing all groups with synchronization being off with the wrong approach.
I understand our problem, but I believe there must be a better solution.

What do you use as value for `groupAdminMemberPattern` in your installation? Which groups from your AD do you want in Gitblit? Could these be grouped under a common OU?

fzs:
I'm closing this pull request as it cannot be merged the way it is as discussed above.
While the problem is valid, this is not the solution. Further discussion of the problem to find a proper solution can be hold in #1156, if desired.

PR #1171, opened 2016-12-30
Update UserManager to support construction of IUserServices with IRuntimeManager as a constructor parameter

pingunaut:
In [gitblit-crowd-plugin](https://github.com/pingunaut/gitblit-crowd) I make use of my own subclass of ConfigUserService.
Unfortunatelly i'm not able to get the data directory from the settings during construction, which is quite bad.

A very simple approach of supporting customized IUserServices is implemented in this PR:
Behavior at the moment:
* Try to instantiate an IUserService with default constructor. 
* If class not found ->create ConfigUserService from conf file
* If instantiation fails -> log error (and die... this will cause NullPointerExceptions afterwards

Newly implemented behavior:
* Try to instantiate an IUserService with default constructor. 
* If class not found -> create ConfigUserService from conf file
* If instantiation with default constructor fails -> Try again with constructor having IRuntimeManager as a parameter
* If this instantiation fails also -> log error (and die... this will cause NullPointerExceptions afterwards

This way no further changes in existing UserServices are needed. New implementations will have the possibility to access all application settings by implementing a constructor like the following:
```java
public class MyCoolUserService implements IUserService {
  public MyCoolUserService(IRuntimeManager manager){
    //do whatever you like. you know the runtimeManager now
  }
}
```

### DIFF
--- a/src/main/java/com/gitblit/IUserService.java
+++ b/src/main/java/com/gitblit/IUserService.java
@@ -26,6 +26,9 @@ import com.gitblit.models.UserModel;
  * Implementations of IUserService control all aspects of UserModel objects and
  * user authentication.
  *
+ * Plugins implementing this interface (which are instantiated during {@link com.gitblit.manager.UserManager#start()}) can provide
+ * a default constructor or might also use {@link IRuntimeManager} as a constructor argument which will be passed automatically then. 
+ *
  * @author James Moger
  *
  */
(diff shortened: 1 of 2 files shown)


gitblit:
I like this idea.  Would you please make instantiation of a user service a new method?  i.e. extract `realmClass.newInstance()` and your catch block into a new method.

Issue #1166, opened 2016-12-14
Use more secure hashes for passwords


fzs:
Gitblit currently uses outdated hash functions for passwords stored in the user file. It should be updated to use modern hashes considered secure.

jblaufuss:
Yes! I just came here to report this same issue.  Looking at users.conf, passwords are hashed using MD5, which is easily cracked.  A modern system should use something like scrypt: https://en.wikipedia.org/wiki/Scrypt

pingunaut:
I opened a pull request to address this issue. It's adding PBKDF2WithHmacSHA256 hash algorithm while being fully backward compatible.

A general topic which should be discussed in my opinion is the following:
When the realm.passwordStorage property is updated, the stored password hashes will only be updated if the user changes his password which might take a very long time.
Would it be better to check if the property was changed on every login? 

Best regards,
Martin

jblaufuss:
I have never actually built and authentication like this, but have have heard it discussed.  IIRC, the "right way" to do a hash algorithm migration is this:

* The password validation function has two cases:

1. Hash the password with the new algorithm and verify the output against the stored hash.

2. If the above fails, then hash the password with the _old_ algorithm, then hash the output with the _new_ algorithm, then verify the chained output against the stored hash.

* Then, on upgrade, automatically hash all the hashes in users.conf with the new algorithm.  They'll be able to be verified by case 2.

pingunaut:
Ah, I wasn't aware of this migration process, but it sounds reasonable 👍. I'll update the PR asap.
I share the opinion that MD5 for password hashing. I'm a little afraid of dropping support for MD5 though. Who knows if someone used it for some integration or something... and there's also the option to use plaintext... don't know what's a good solution here :/

fzs:
The password hash to be used is configurable. That means that it could have changed over time and that multiple different schemes are in use in the users.conf file. There is no way to know what the "old" algorithm was, I think. I believe the suggested process works for applications with a fixed hashing algorithm that was changed, not in this case.

pingunaut:
@fzs you are right. I tried to implement the above suggested migration process and kept the information about previous algorithms in an additional prefix. But it turned out to get really complex and not very maintainable :/

fzs:
I don't think it is much of a problem and needs to be solved automatically. A Gitblit administrator can decide if he is fine with leaving old hashed values as they are or to what degree she wants to bug her users to update their passwords so that they are hashed with a new algorithm.
[...]

PR #1172, opened 2017-01-01
Added possibility to use secure hashes to store passwords

pingunaut:
Addresses #1166

PBKDF2WithHmacSHA256 is used instead of MD5.
https://www.owasp.org/index.php/Hashing_Java was used as a reference for the implementation

### DIFF
--- a/src/main/distrib/data/defaults.properties
+++ b/src/main/distrib/data/defaults.properties
@@ -854,12 +854,14 @@ realm.userService = ${baseFolder}/users.conf
 realm.authenticationProviders =
 
 # How to store passwords.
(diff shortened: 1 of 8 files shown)


fzs:
With which setup did you test this? OS, Java vendor and version?

pingunaut:
Linux (Fedora), openjdk-8
I am able to test against Windows7, Oracle JRE 7 later this day

fzs:
Okay, thanks.
I have started to integrate your changes, but will have/want to rewrite
them some.

pingunaut:
Not as soon as I thought, but finally verified: Windows 7/Oracle JRE 7 is woking fine

Issue #1132, opened 2016-09-22
Payara Micro 4.1 (163) throws SAXParseException


Quix0r:
With the micro version deployment (and also cluster-integration) is very easy:

> $ java -jar payara-micro-4.1.1.163.jar --deploy gitblit-1.8.0.war

Sadly gitblib (downloaded from main website) ends up this way:

```
[2016-09-22T10:30:08.990+0200] [Payara Micro 4.1] [WARNING] [] [] [tid: _ThreadID=1 _ThreadName=main] [timeMillis: 1474533008990] [levelValue: 900] [[
  Error occurred
org.xml.sax.SAXParseExceptionpublicId: jar:file:/home/payara/payara-micro-4.1.1.163.jar!/schemas/web-app_3_0.xsd; lineNumber: 8; columnNumber: 27; Deployment descriptor file WEB-INF/web.xml in archive [gitblit-1.8.0.war].  TargetNamespace.1: Expecting namespace 'http://java.sun.com/xml/ns/j2ee', but the target namespace of the schema document is 'http://java.sun.com/xml/ns/javaee'.
        at com.sun.enterprise.deployment.io.DeploymentDescriptorFile.read(DeploymentDescriptorFile.java:328)
        at com.sun.enterprise.deployment.io.DeploymentDescriptorFile.read(DeploymentDescriptorFile.java:245)
        at com.sun.enterprise.deployment.archivist.Archivist.readStandardDeploymentDescriptor(Archivist.java:666)
        at com.sun.enterprise.deployment.archivist.ApplicationFactory.createApplicationFromStandardDD(ApplicationFactory.java:210)
        at org.glassfish.javaee.core.deployment.DolProvider.getNameFor(DolProvider.java:269)
        at org.glassfish.internal.deployment.GenericHandler.getDefaultApplicationName(GenericHandler.java:152)
        at org.glassfish.deployment.admin.DeployCommand.preAuthorization(DeployCommand.java:262)
        at com.sun.enterprise.admin.util.CommandSecurityChecker$1.run(CommandSecurityChecker.java:187)
        at com.sun.enterprise.admin.util.CommandSecurityChecker$1.run(CommandSecurityChecker.java:183)
        at java.security.AccessController.doPrivileged(Native Method)
        at javax.security.auth.Subject.doAs(Subject.java:356)
        at com.sun.enterprise.admin.util.CommandSecurityChecker.authorize(CommandSecurityChecker.java:183)
[...]

PR #1178, opened 2017-01-16
Update to web.xml, fix to #1132

srbala:
Update to web.xml, fix to #1132

Fixes to namespace to fix xml parse error, where strict validation required


fzs:
This PR fixes PR #275 which was actually a botched update to Servlet 3.0 as it only updated the version number in the file name, but not the path or the version in the declaration. It will drop Tomcat 6 support, which might not be a problem, given that Tomcat 6 was EOL'ed in 2016.

The `weblogic.xml` file still lists the 2.4 web-app, though. Given that it was written for Weblogic 12.1.1, which implements Java EE 6, could you also update the `weblogic.xml` file with the new path/version to keep the two files consistent?

srbala:
weblogic.xml file is updated now

fzs:
Thank you.

PR #1183, opened 2017-01-23
Documentation  update : "behind apache" also edit "Ajax-Location" header

RainerW:
In case apache already terminates https and gitblit only receives http.  Using the "New Ticket" function will respond with an "http://..." in the "Ajax-Location" header. So i added the needed "Header edit ...." line to the documentation.


gitblit:
Good catch!

PR #1185, opened 2017-01-24
Sending email (certificate zip file) based on locale that user selected

WilliamFromTW:
create i18n file in directory src\main\distrib\data
1.  mail.tmpl_xx_YY
2.  instructions_xx_YY

new paramter in  file src\main\java\com\gitblit\wicket\GitBlitWebApp.properties
gb.emailClientCertificateSubject = Your Gitblit client certificate for {0}


WilliamFromTW:
failed if user.getPreferences().getLocale() is null

Issue #1184, opened 2017-01-24
RPC LIST_BRANCHES for a given repository


mystygage:
The RPC LIST_BRANCHES lists all branches from all repositories. This produces a lot of overhead, especially in a gitblit instance with a bunch of repositries and a lot of branches.

It would be useful to have a RPC request like LIST_REPOSITORY_BRANCHES for a given repository.

PR #1186, opened 2017-01-24
New rpc endpoint for listing all branches from a given repository

mystygage:
fix issue #1184


gitblit:
See the note [here](https://github.com/mystygage/gitblit/blob/c4a9e435e34ef7e9bd2ac5c71a96855502091579/src/main/java/com/gitblit/Constants.java#L410-L411).

What if we instead added support to the existing `LIST_BRANCHES` call for a single repo?

mystygage:
> What if we instead added support to the existing `LIST_BRANCHES` call for a single repo?

Ok, I'll rewrite my code.

gitblit:
No, leave it.  You've moved the constant, that solves the immediate 
problem.

gitblit:
I merged and then reverted.  We'll need a new PR to get this in.  We forgot to enforce access controls on the repo.

mystygage:
Ok. So modifying `LIST_BRANCHES` with existing access control seems like the best solution

fzs:
New PR is #1192

PR #1192, opened 2017-01-26
List branches from only one repository if a repository name is given

### LINKED ISSUE #1184: RPC LIST_BRANCHES for a given repository

mystygage:
Fix #1184 

I'm not sure whether this change should increase the rpc version number or not...

### DIFF
--- a/src/main/java/com/gitblit/servlet/RpcServlet.java
+++ b/src/main/java/com/gitblit/servlet/RpcServlet.java
@@ -133,6 +133,10 @@ public class RpcServlet extends JsonServlet {
 							model.name));
 					continue;
 				}
+				if(!StringUtils.isEmpty(objectName) && !objectName.equals(model.name)) {
+					// skip repository if a name was submitted and it doesn't match
+					continue;
+				}
 				// get local branches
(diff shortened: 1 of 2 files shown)


src/main/java/com/gitblit/servlet/RpcServlet.java
```diff
@@ -133,6 +133,10 @@ protected void processRequest(HttpServletRequest request, HttpServletResponse re
 							model.name));
 					continue;
 				}
+				if(objectName != null && !objectName.equals(model.name)) {
```
gitblit: `if (!StringUtils.isEmpty(objectName) && !objectName.equals(model.name)) {`

mystygage: Done

gitblit:
Thanks!

PR #1194, opened 2017-01-28
Escape folder names in JGitUtils to allow special characters. #999

### LINKED ISSUE #999: special chars in folders name
I have a folder named "[logs]". When I try to open it with gitblit, it returns:

"Internal error"

How can I solve it?

thank you

pingunaut:
```String.replaceFirst(String regex, String replacement)``` works with regex.

If a folder name contains chars which are parsed by regex, the replaceFirst will not work as expected.
E.g: if there is a pathString with value "[foo]/bar" and a path having the value "[foo]", then the expected behavior is to result in a string with value "bar".
Fortunately the path can easily be escaped using ```Pattern.quote() ```

### DIFF
--- a/src/main/java/com/gitblit/wicket/pages/EditUserPage.java
+++ b/src/main/java/com/gitblit/wicket/pages/EditUserPage.java
@@ -27,7 +27,6 @@ import org.apache.wicket.extensions.markup.html.form.palette.Palette;
 import org.apache.wicket.markup.html.form.Button;
 import org.apache.wicket.markup.html.form.CheckBox;
 import org.apache.wicket.markup.html.form.Form;
-import org.apache.wicket.markup.html.form.IChoiceRenderer;
 import org.apache.wicket.markup.html.form.TextField;
 import org.apache.wicket.model.CompoundPropertyModel;
 import org.apache.wicket.model.Model;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/main/java/com/gitblit/wicket/pages/EditUserPage.java [hunk 1]
```diff
@@ -106,8 +106,21 @@ protected void setupPage(final UserModel userModel) {
 		final List<RegistrantAccessPermission> permissions = app().repositories().getUserAccessPermissions(userModel);
 
 		final Palette<String> teams = new Palette<String>("teams", new ListModel<String>(
-				new ArrayList<String>(userTeams)), new CollectionModel<String>(app().users()
-				.getAllTeamNames()), new StringChoiceRenderer(), 10, false);
+				new ArrayList<String>(userTeams)), new CollectionModel<String>(app().users().getAllTeamNames()),
```
gitblit: Assuming this change is necessary, I would expect we'd need it anywhere we use a palette for teams.

pingunaut:
I reverted the code on UserEditPage for now, so the pull request only addresses #999 
As the other issue might need further changes, it's better to handle it within a seperate PR

fzs:
I cherry-picked the one commit that fixes #999. (3bff35cd)
Closing this, considering it merged.

PR #1201, opened 2017-02-15
Fix #1114 HttpUtils getGitblitURL does not support nonstandard ports

### LINKED ISSUE #1114: HttpUtils getGitblitURL does not support nonstandard ports
We are using reverse proxy. Gitblit is running only HTTP on port 29080 and apache does reverse proxy to HTTPS 4000. 

It is caused by regression in pull request #1086, the solution of #1083.

It is because the port number is doubled. In the path, there is hardcoded check for 80 and 443, but there is no check if host already contains the port number.

With `web.canonicalUrl = https://ourhost:4000` graph image is generated as

```
<img src="https://ourhost:4000:4000/graph/?r=........"></img>
```

`web.canonicalUrl` should contain nonstandard port.
Setting just `web.canonicalUrl = https://ourhost` also does not work.

j123b567:
X-Forwarded-Host can contain port number and it is added twice in that situation
This fix just prevent adding port number if it is already there

### DIFF
--- a/src/main/java/com/gitblit/utils/HttpUtils.java
+++ b/src/main/java/com/gitblit/utils/HttpUtils.java
@@ -110,7 +110,9 @@ public class HttpUtils {
 		sb.append(host);
 		if (("http".equals(scheme) && port != 80)
 				|| ("https".equals(scheme) && port != 443)) {
-			sb.append(":").append(port);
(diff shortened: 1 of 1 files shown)


gitblit:
Great, thank you!

PR #1202, opened 2017-02-20
Prevent last column on Tree page from wraping

j123b567:
With localizations, there can be longer text then 13em so there is a line break.
This fix prevent this line breaking and thus every line has normal height again.


gitblit:
Thanks!

PR #1205, opened 2017-02-28
introduces writeSignoffCommit and requireScore features

fuero:
This pull request introduces 2 features to Gitblit:

**writeSignoffCommit**:
When using the merge button on a ticket, all reviewers are added in a 'Signed-off-by: reviewer <reviewer@example.com>' commit message line. Can be enabled/disabled in the same manner
as requireApproval.
Defaults to being disabled.

**requireScore**:
Allows specifying a particular score a ticket should have (i.e. a number of people who have to sign off on it) before the merge button appears in the web ui. Configurable per repo, requires the requireApproval setting.
Defaults to being disabled.

I hope the approach is clean enough and that other people want this as well.

This would save us from setting up Gerrit :-)


src/main/java/com/gitblit/git/PatchsetReceivePack.java
```diff
@@ -1268,7 +1268,28 @@ private void updateReflog(RefUpdate ru) {
...
+		StringBuilder messageBuilder = new StringBuilder();
+		messageBuilder.append(MessageFormat.format("Merged #{0,number,0} \"{1}\"", ticket.number, ticket.title));
+
+		// Add Signed-off-by tags to commit message footer if there are reviewers on this patchset
+		// and the setting is enabled on this repo.
+		if (repository.writeSignoffCommit) {
+			messageBuilder.append("\n\n");
+			for (Change change : ticket.getReviews(patchset)) {
+				UserModel ruser = gitblit.getUserModel(change.author);
+				messageBuilder.append(MessageFormat.format(
```
gitblit: This looks like it would add a `sign-off` for someone who gave a negative review but the total score exceeded the minimum requirement.  Am I reading that wrong?

fuero: Nope, you're correct. Have a look at the follow-up commit, this should fix it.

gitblit:
@fuero I like this proposal.  Do you consider it feature complete?

gitblit:
I wonder if `Reviewed-by` would be more appropriate, see [conventions](https://git.wiki.kernel.org/index.php/CommitMessageConventions).

fzs:
>
>
> I wonder if Reviewed-by would be more appropriate, see conventions
> <https://git.wiki.kernel.org/index.php/CommitMessageConventions>.
>

Agreed. Or make it configurable what the line is, if you need it to be
Signed-off in your organisation.

fuero:
As I understand it, the definition of `sign-off` is [somewhat](https://gerrit-review.googlesource.com/Documentation/user-signedoffby.html) [fluent](https://github.com/wking/signed-off-by).

Another setting containing the header name to be added could be introduced.

The actual meaning of the `sign-off` setting must be defined by the people relying on it.

As for it being feature complete: I've introduced this to meet our internal needs, which are evolving as well.

fzs:
Aren't both sources agreeing, that `Signed-off-by` means that the signer has authorship of the code, while `reviewed-by` is for people reviewing?

Anyhow, as for an additional setting, I'd personally rather vote for using the existing setting, and if it has no value, nothing is added. I'm a friend of controlling setting proliferation.

Maybe the documentation should mention that this is only valid if a merge commit is created, or is that obvious enough for everyone?

fuero:
@fzs I've adapted my code to reflect your idea. Is there a way to add the conventions link to message displayed in the GUI? It didn't seem to like HTML tags there.

fzs:
That should work fine. But it is better to keep the link in code so that it can be changed easier and allows for easier translation. See commit 8130906c48dcd4886bb8b28a6c1055489b499140.

fuero:
Hope this is ok, it does stick out from the other options now though.

fuero:
rebased to match current origin master

PR #1207, opened 2017-03-03
Better Czech localization

j123b567:
After using Czech translation for a while, my colegues give me some feedback. The main difference is how to name a "repository".
There are also some more fixes in spelling of some words.


gitblit:
Fantastic!  Thank you.

PR #1209, opened 2017-03-13
Change version to properties

marcomsousa:
Raise dependencies numbers.

This update raise all dependecies to the last version, some of them will be an major change, and will break this project.
Please take some time to do this, because you are using (2014) old version (bugs, security problems, etc..)

If you see that this will take too much time, you can only upgrade the easy ones at this moment.
Lots of them are only fixes or compatible changes. And shouldn't break..

Like: jetty, slf4j, jgit, groovy, bouncycastle, sshd, mina, guice, gson, ivy, commons-compress, commons-io, freemarker, commons-codec, jedis, jacoco

### DIFF
--- a/build.moxie
+++ b/build.moxie
@@ -103,20 +103,41 @@ repositories: central, eclipse-snapshots, eclipse, gitblit
 
 # Convenience properties for dependencies
 properties: {
-  jetty.version  : 9.2.13.v20150730
-  slf4j.version  : 1.7.12
-  wicket.version : 1.4.22
-  lucene.version : 4.10.4
-  jgit.version   : 4.1.1.201511131810-r
-  groovy.version : 2.4.4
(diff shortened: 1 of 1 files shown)


marcomsousa:
I don't want to be bad interpreted, I know that is lot's of work.. I just want to be reflected here..

fzs:
This would probably better have been an issue, instead of a pull request. As you said, this is a lot of work that needs a lot of testing. My suspicion is that you didn't even compile that, as I believe it wouldn't compile. A few of the updates cause incompatibilities. There is even a whole branch dedicated to updating Wicket. So there probably is zero chance of this PR getting merged.

gitblit:
Keeping up-to-date is a reasonable request but the proposed PR is an unreasonable way to achieve that.  :)

PR #1216, opened 2017-03-18
Update sshd to 1.2.0 and mina to 2.0.10

paladox:


paladox:
I haven't really tested if ssh still works but it builds ok with this patch.

fzs:
I do hope you realise that updating a dependency involves a lot of testing to make sure that everything still works as before, and is not done with just making things compile.

PR #1219, opened 2017-03-18
Bump min java support to 1.8

### LINKED ISSUE #1218: Bump min java version to 1.8
We should bump the min java version to 1.8 in the next release so we can bump jetty version and other deps and other deps which requires java 1.8+.

paladox:
Fixes #1218


flaix:
This was manually merged after rebasing the branch and adding more changes to switch to Java 8. Unfortunately Github did not pick up on this so I have to close it now and it doesn't show as merged.

PR #1232, opened 2017-05-02
Adds Docker Release Build, suitable for dockerhub automatic builds

goettl79:
.) Adds Docker Release Build and pre_commit hook for docker hub / ant build
.) Configured auto build with dockerhub, 4 my own account. Proposal is to create an official "gitblit" account and build it there.
.) Test container with "docker run -p 8080:8080 -it goettl/gitblit"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,32 @@
+FROM tomcat:9.0
+ENV GITBLIT_HOME=/opt/gitblit-data
+
+# Prepare Environment
+RUN groupadd git \
+&& useradd -g git git \
+&& apt-get --yes update \
+&& apt-get --yes install redis-server \
+&& apt-get clean \
+&& mkdir -p /opt/gitblit-data \
+&& rm -r /usr/local/tomcat/webapps/*
+
+# Install Gitblit
+COPY ./build/target/*.war  /usr/local/tomcat/webapps/ROOT/
+RUN cd /usr/local/tomcat/webapps/ROOT/ &&  unzip  *.war && rm *.war \
+&& mv /usr/local/tomcat/webapps/ROOT/WEB-INF/data/* /opt/gitblit-data \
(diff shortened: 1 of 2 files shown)


fzs:
I'm not sure I understood the idea behind this correctly. I take it, it was not to get a CI build, but to deploy "nightlies" as container images so that they are available on Docker Hub?

goettl79:
Yes, first step towards this back then. However now a bit out of date 😉

Florian Zschocke <notifications@github.com> schrieb am Do., 6. Feb. 2020,
22:47:

fzs:
Well, I don't know about "first step". I mean, the idea is not necessarily out of date.
So for me the question is, is there still interest in providing nightly builds (in general, and) as Docker containers on Docker Hub. No matter where they are built and and how they get there, would that be useful to enough people to get it set up.

(I have the general problem of knowing how large a user base or user interest Gitblit still has. It's hard to find out without some way to count the installed base.)

goettl79:
If it makes still sense to have i think it might also help to see how big the user base still is. All pulls from docker hub would be visible... Most reasonable way would be to become a official image. See here https://docs.docker.com/docker-hub/official_images/

flaix:
I am closing this PR as done, after having made nightly builds of Gitblit available on Docker Hub now.

I opted for a different approach, making use of the gitblit-docker repository. With commit gitblit/gitblit-docker@cfd130ae1edde96f1547578b37fcb2a1e574da78 a Dockerfile for a snapshot version can be created. With a scheduled workflow in my fork of gitblit, a nightly version is build (if there were changes) and pushed to Docker Hub under the `gtiblit/gitblit:nightly` tag. I have to do it from my fork for now as login data for [Docker Hub](https://hub.docker.com/r/gitblit/gitblit) is needed and I cannot add them to the official gitblit/gitblit repo.

Issue #1237, opened 2017-05-20
Unable to view summary/tickets page


drognisep:
[Originally posted on super**user**](https://superuser.com/questions/1210991/gitblit-war-wicketruntimeexception-on-any-repository-summary-page)

I'm currently unable to view the summary page on any repository, or view tickets. Please see the super**user** link above for detailed information.

Here's the stack trace for attempting to load a summary page, specifically for **jPortfolio**:
```
2017-05-19 18:36:34 [ERROR] Can't instantiate page using constructor public com.gitblit.wicket.pages.SummaryPage(org.apache.wicket.PageParameters) and argument r = "jPortfolio.git"
org.apache.wicket.WicketRuntimeException: Can't instantiate page using constructor public com.gitblit.wicket.pages.SummaryPage(org.apache.wicket.PageParameters) and argument r = "jPortfolio.git"
        at org.apache.wicket.session.DefaultPageFactory.createPage(DefaultPageFactory.java:212)
        at org.apache.wicket.session.DefaultPageFactory.newPage(DefaultPageFactory.java:89)
        at org.apache.wicket.request.target.component.BookmarkablePageRequestTarget.newPage(BookmarkablePageRequestTarget.java:305)
        at org.apache.wicket.request.target.component.BookmarkablePageRequestTarget.getPage(BookmarkablePageRequestTarget.java:320)
        at org.apache.wicket.request.target.component.BookmarkablePageRequestTarget.processEvents(BookmarkablePageRequestTarget.java:234)
        at org.apache.wicket.request.AbstractRequestCycleProcessor.processEvents(AbstractRequestCycleProcessor.java:92)
        at org.apache.wicket.RequestCycle.processEventsAndRespond(RequestCycle.java:1279)
        at org.apache.wicket.RequestCycle.step(RequestCycle.java:1358)
        at org.apache.wicket.RequestCycle.steps(RequestCycle.java:1465)
        at org.apache.wicket.RequestCycle.request(RequestCycle.java:545)
        at org.apache.wicket.protocol.http.WicketFilter.doGet(WicketFilter.java:486)
        at org.apache.wicket.protocol.http.WicketFilter.doFilter(WicketFilter.java:319)
        at com.google.inject.servlet.FilterChainInvocation.doFilter(FilterChainInvocation.java:88)
        at com.gitblit.servlet.EnforceAuthenticationFilter.doFilter(EnforceAuthenticationFilter.java:99)
[...]

PR #1238, opened 2017-05-21
Add nullcheck  during favorite protocol determination

pingunaut:
Add nullcheck  during favorite protocol determination. fix for #1237


gitblit:
Thanks!

PR #1243, opened 2017-05-31
Fixes the 30 seconds timeout when cloning over HTTP

sebastianopilla:
Adds the

`server.httpTimeout`

setting to allow cloning big repositories over HTTP. This change fixes the java.util.concurrent.TimeoutException issue described in https://groups.google.com/d/topic/gitblit/UvDC48NpmF4/discussion .

### DIFF
--- a/src/main/distrib/data/defaults.properties
+++ b/src/main/distrib/data/defaults.properties
@@ -2134,6 +2134,13 @@ server.requireClientCertificates = false
 # RESTART REQUIRED
 server.shutdownPort = 8081
 
+# Timeout (in milliseconds) for http and https requests
+# Increase this value if you get java.util.concurrent.TimeoutException errors
+#
+# SINCE 1.9.0
+# RESTART REQUIRED
+server.httpTimeout = 30000
+
 #
 # Gitblit Filestore Settings
 #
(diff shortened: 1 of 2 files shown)


src/main/distrib/data/defaults.properties
```diff
@@ -2134,6 +2134,13 @@ server.requireClientCertificates = false
 # RESTART REQUIRED
 server.shutdownPort = 8081
 
+# Timeout (in milliseconds) for http and https requests
+# Increase this value if you get java.util.concurrent.TimeoutException errors
+#
+# SINCE 1.9.0
+# RESTART REQUIRED
+server.httpTimeout = 30000
```
gitblit: How about `server.httpIdleTimeout`

src/main/distrib/data/defaults.properties
```diff
@@ -2134,6 +2134,13 @@ server.requireClientCertificates = false
 # RESTART REQUIRED
 server.shutdownPort = 8081
 
+# Timeout (in milliseconds) for http and https requests
```
gitblit: How about `Http idle timeout (in milliseconds)...`

sebastianopilla:
I've made the changes, I can't figure out if there's anything else that I need to do to have them applied to the pull request however. Let me know if I need to drop this pull request and create a new one.

gitblit:
Thanks!

PR #1252, opened 2017-07-29
Update new jgit release (4.8.0.201706111038-r)

jvanhercke:
Current release of gitblit is based on version 4.1.1.201511131810-r
which fails to start throwing an UnsupportedCharsetException if a
commit contains an unknown character set encoding (e.g. System).

This new jgit version recovers from the exception.

See changes to class org.eclipse.jgit.util.RawParseUtils,

public static Charset parseEncoding(final byte[] b)


gitblit:
The more recent releases of JGit are also the cause of performance regressions in Gerrit due to autogc.  I haven't really investigated it myself, that is the scuttlebutt on the JGit mailing list.

As for the charset issue, I did run into that a long while ago.  My solution at the time was to hack the fix into JGit using reflection.  Perhaps you could determine the offending charset and map it [here](https://github.com/gitblit/gitblit/blob/master/src/main/java/com/gitblit/manager/RepositoryManager.java#L1948-L1960).

jvanhercke:
Ok, I understand. I will look into it. Nice hack b.t.w., but unfortunately it does not prevent the system from crashing.

PR #1253, opened 2017-07-29
Add wrapper class to return a default encoding

jvanhercke:
Unknown encodings may cause gitblit to fail to start. This modification
injects a wrapper class in the JGit internal to fake a valid return value.


flaix:
This has been superseded by commit e8a7b2d9f4a7ead8a195158ac8dc233735e00858. I integrated the commit in a branch in gitblit, to keep it for reference.
But the update of JGit fixed the issue, plus a fix to Bugtraq which could also trigger an exception, so the complete extra hack was removed.
Closing this as obsolete. Thank you very much for the contribution nevertheless!

Issue #1262, opened 2017-10-20
gitblit can not run on Java-9


godsay1983:
D:\gitblit-1.8.0>gitblit
java.lang.ClassCastException: java.base/jdk.internal.loader.ClassLoaders$AppClassLoader cannot be cast to java.base/java.net.URLClassLoader
        at org.moxie.MxLauncher.addJarFile(MxLauncher.java:170)
        at org.moxie.MxLauncher.main(MxLauncher.java:117)
java.lang.ClassCastException: java.base/jdk.internal.loader.ClassLoaders$AppClassLoader cannot be cast to java.base/java.net.URLClassLoader
        at org.moxie.MxLauncher.addJarFile(MxLauncher.java:170)
        at org.moxie.MxLauncher.main(MxLauncher.java:117)
java.lang.ClassCastException: java.base/jdk.internal.loader.ClassLoaders$AppClassLoader cannot be cast to java.base/java.net.URLClassLoader
        at org.moxie.MxLauncher.addJarFile(MxLauncher.java:170)
        at org.moxie.MxLauncher.main(MxLauncher.java:117)
java.lang.ClassCastException: java.base/jdk.internal.loader.ClassLoaders$AppClassLoader cannot be cast to java.base/java.net.URLClassLoader
        at org.moxie.MxLauncher.addJarFile(MxLauncher.java:170)
        at org.moxie.MxLauncher.main(MxLauncher.java:117)
java.lang.ClassCastException: java.base/jdk.internal.loader.ClassLoaders$AppClassLoader cannot be cast to java.base/java.net.URLClassLoader
        at org.moxie.MxLauncher.addJarFile(MxLauncher.java:170)
        at org.moxie.MxLauncher.main(MxLauncher.java:117)
java.lang.ClassCastException: java.base/jdk.internal.loader.ClassLoaders$AppClassLoader cannot be cast to java.base/java.net.URLClassLoader
        at org.moxie.MxLauncher.addJarFile(MxLauncher.java:170)
        at org.moxie.MxLauncher.main(MxLauncher.java:117)
java.lang.ClassCastException: java.base/jdk.internal.loader.ClassLoaders$AppClassLoader cannot be cast to java.base/java.net.URLClassLoader
        at org.moxie.MxLauncher.addJarFile(MxLauncher.java:170)
        at org.moxie.MxLauncher.main(MxLauncher.java:117)
[...]

PR #1266, opened 2017-11-03
Added javax.activation dependency "

pingunaut:
thanks to this dependency, one can start gitblit with java 9 without using deprecated internal module --add-modules java.activation

addresses #1262


fzs:
I don't see this fixing the mentioned #1262. It is a necessary fix to run on Java9+, but the error in #1262 comes from the `MxLauncher` having problems with Java 9+. That will need a different fix.

PR #1224, opened 2017-04-08
collapsible group repositories

ybosy:
This resolves #527 
If you have any suggestions or comments, please let me know.

Issue #725, opened 2015-08-12
Feature request: Support nested groups on the Repositories page


gitblit:
Originally reported on Google Code with ID 429

```
In my organization we have a lot of Git repositories and would like to be able to group
the repositories in a nested fashion, e.g.

Managers
- Manager X
  - Server
  - Client
- Manager Y
  - Server
  - Database
  - Client
etc.

Here https://code.google.com/p/gitblit/issues/detail?id=406#c1 it is mentioned that
it's possible to nest repos with '/', so what I'm suggesting is really just that the
interface reflects this tree structure more clearly than now. Furthermore the GUI change
here https://code.google.com/p/gitblit/issues/detail?id=231&q=collapse#makechanges,
will fit nicely into this feature.


```

Reported by `jakob@jboysen.dk` on 2014-05-20 11:04:35

fzs:
Google Code issue 406 was imported as issue #702.
Google Code issue 231 was imported as issue #527.

fzs:
This is implemented by PR #1267

Issue #527, opened 2015-08-12
GUI: expand and flip group repositories


gitblit:
Originally reported on Google Code with ID 231

```
What steps will reproduce the problem?
1. Have multiple repository which are grouped
2. All repositories and repository groups are rendered in one table
3. When showing huge amounts of repositories the table gets confusing.

What is the expected output? What do you see instead?
I would like to have the possibility to flip and expand the grouped repositories (table
rows). It would be nice to have a application property to set a default value for the
grouped view (expanded / flipped).

What version of the product are you using? On what operating system?
v1.2.1 GitBlit WAR on Apache Tomcat 7 on Debian 4.6

Please provide any additional information below.
see the screenshots:
current_view.png -> this is the current dashboard view with grouped repositories
new_view.png -> this is a possible resolution for the requirement; using [+] and [-]
in front of the group name to switch between expanded and flipped in view.

link:
howto for flipping table rows: http://snippets.fiae.ws/show/11/
```

Reported by `stephan.krull.ecg` on 2013-04-15 09:20:51

-----
- _Attachment: current_view.png<br>![current_view.png](https://storage.googleapis.com/google-code-attachments/gitblit/issue-231/comment-0/current_view.png)_
- _Attachment: new_view.png<br>![new_view.png](https://storage.googleapis.com/google-code-attachments/gitblit/issue-231/comment-0/new_view.png)_

fzs:
Closed by PR #1224

PR #1267, opened 2017-11-06
Add support nested groups on the Repositories page

pingunaut:
Fix for #725.

This includes #1224 from @ybosy, because a lot of the code of it was reusable.

Also covers  #527 

image

(Panel contains a lot of copy-paste code from RepositoriesPanel :( But this way I was able to keep the original RepositoriesPanel without changing it a lot. ...this could be done a lot easier after a wicket upgrade in my opinion ;))


YandexWillKillYou:
Just got a task to realize similar functionality. Nice to see it done!

fzs:
What would you think about listing repositories first?

Screenshot 2019-06-15 at 16 36 32

Leaving out the "guard rails" would look cleaner, but with many folders it might get difficult to see what group a repository belongs to. On the other hand, this is the opposite sorting to what one is used from a file explorer.

pingunaut:
I like your approach a lot. looks much better in my opinion :+1:

fzs:
Closed by a squash merge to master in commit 0bace7e

PR #1272, opened 2018-01-03
Better compatibility with SSH clients (eg, Jenkins/BlueOcean) and add EdDSA key support by upgrading sshd

rpardini:
Hello! Two commits here.
First one is the minimum needed to make Gitblit work with Jenkins/BlueOcean (resolves the "MAC corrupt" exception). It upgrades sshd to 1.2.0 with no further changes.
Second one upgrades sshd to 1.6.0, and adds a dependency on eddsa, and adds code to register this provider, which in fact (tested) adds support for my new ED25519 ssh key. It also upgrades JGit to a newer point release which is supposed to fix bugs only.
I'm running two production instances of Gitblit with these patches, and finally enjoying Gitblit + Jenkins/Blueocean pipelines.

### DIFF
--- a/build.moxie
+++ b/build.moxie
@@ -107,12 +107,12 @@ properties: {
   slf4j.version  : 1.7.12
   wicket.version : 1.4.22
   lucene.version : 5.5.2
-  jgit.version   : 4.1.1.201511131810-r
+  jgit.version   : 4.1.2.201602141800-r
   groovy.version : 2.4.4
   bouncycastle.version : 1.52
   selenium.version : 2.28.0
   wikitext.version : 1.4
-  sshd.version: 1.2.0
+  sshd.version: 1.6.0
   mina.version: 2.0.16
   guice.version : 4.0
   # Gitblit maintains a fork of guice-servlet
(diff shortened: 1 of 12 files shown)


rpardini:
There's a Jenkins issue that would be resolved with this. [JENKINS-45769](https://issues.jenkins-ci.org/browse/JENKINS-45769) -- I did not dare try to fix this on Jenkins side.

rpardini:
Some proof of EdDSA support...
On the web interface:
screen shot 2018-01-03 at 16 28 55

Via SSH "keys which"...
screen shot 2018-01-03 at 16 28 30

dbywalec:
Though I do not use ED25519, I have merged your commits to our Gitblit server and it works. I have been missing this merge work for a long time. You made my day. Good work. Thank you.

flaix:
I'm reopening this pull request since only one commit was merged and the support for eddsa is still pending.

build.moxie
```diff
@@ -107,13 +107,13 @@ properties: {
...
   lucene.version : 5.5.2
-  jgit.version   : 4.1.1.201511131810-r
+  jgit.version   : 4.1.2.201602141800-r
   groovy.version : 2.4.4
   bouncycastle.version : 1.52
   selenium.version : 2.28.0
   wikitext.version : 1.4
-  sshd.version: 1.0.0
-  mina.version: 2.0.9
+  sshd.version: 1.6.0
```
bturner, line 115: Apache SSHD 1.6.0 is still extremely old ([tagged June, 2017](https://github.com/apache/mina-sshd/releases/tag/sshd-1.6.0)) and has publicly documented security issues. The current release is 2.8.0.

It may be an upgrade you want to tackle separately, but it's likely worthwhile to do so. In particular, going to SSHD 2.8 will have [built-in support for rsa-sha2-256 and rsa-sha2-512](https://issues.apache.org/jira/browse/SSHD-1216) which is enabled by default. Since OpenSSH 8.8 disables ssh-rsa by default, this will save your end users a lot of hassle.

rpardini: Yeah when I sent this PR (in 2018) life was different. I tried to close this PR but failed. Good luck!

PR #1277, opened 2018-02-03
Added tika text extraction support for lucene indexing 

mitjale:
I've added text extraction support for Lucene indexing from Apache Tika parsers, thus enabling search on pdf and office documents

### DIFF
--- a/build.moxie
+++ b/build.moxie
@@ -180,6 +180,7 @@ dependencies:
 - compile 'redis.clients:jedis:2.6.2' :war
 - compile 'ro.fortsoft.pf4j:pf4j:0.9.0' :war
 - compile 'org.apache.tika:tika-core:1.5' :war
+- compile 'org.apache.tika:tika-parsers:1.5' :war
 - compile 'org.jsoup:jsoup:1.7.3' :war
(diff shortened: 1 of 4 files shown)


flaix (changes requested):
I am pretty sure that not every line of the class `LuceneService` changed. It looks like you may have changed indentation. This makes it impossible to review changes.

Never change indentation, line endings, whitespace and such when making functional changes. Always keep functional changes separate and reviewable.

TomaszSzt (changes requested):
I can see You noticed that passing byte[] data can crash server if large file is to be processed. Good.

Yet still You do return String from extractText which still, as You don't limit the size of input file to be processed, opens a way for users to abuse it and use it to crash server by supplying large files enough to cause OutOfMemory exception. There is no problem in supplying 100MB or more PDF.

Either use streaming tika or restrict the number of bytes passsed to tika by using an input stream wrapper. Make the limit configurable if You decide on it.

I'm too missing this functionality, because it is a critical step which prevents GitBlit from being Knowledge Management System for non-coders, but this implementation will crash server with OutOfMemory exception in a very unpredictable way.

Issue #1275, opened 2018-01-29
JUnit tests fail due to missing hello-world.git on GitHub


dbywalec:
I have recently discovered that repository "https://github.com/git/hello-world.git" referenced in JUnit tests had disappeared from GitHub and does not exist anymore.

For this reason a number of JUnit tests fail.

```
	public static void setUp() throws Exception {
		startGitblit();

		if (REPOSITORIES.exists() || REPOSITORIES.mkdirs()) {
			cloneOrFetch("helloworld.git", "https://github.com/git/hello-world.git");
			cloneOrFetch("ticgit.git", "https://github.com/schacon/ticgit.git");
			cloneOrFetch("test/jgit.git", "https://github.com/eclipse/jgit.git");
			cloneOrFetch("test/helloworld.git", "https://github.com/git/hello-world.git");
			cloneOrFetch("test/ambition.git", "https://github.com/defunkt/ambition.git");
			cloneOrFetch("test/gitective.git", "https://github.com/kevinsawicki/gitective.git");

			showRemoteBranches("ticgit.git");
			automaticallyTagBranchTips("ticgit.git");
			showRemoteBranches("test/jgit.git");
			automaticallyTagBranchTips("test/jgit.git");
		}
	}
```

Could not all repositories used in tests be forked into "gitblit" account on GutHub in order to make them independent?
This is probably a big simile towards James, the author and owner. :smile:

fzs:
Either that, or have test repositories zipped and checked in to the project as a test resource. I would actually see benefit in the latter as it allows for running tests in an environment where you don't have access to GitHub repos, and you also always have a well defined state of the repository used for tests, which also enabled running tests writing to the repository.

rnveach:
This is similar to https://github.com/gitblit/gitblit/issues/1126  .

I recommend creating repositories at runtime.
Example: https://github.com/rnveach/gitblit/commit/d4b7fecc592f4ca80f0ba99fbbf019907dea7bac#diff-40dc38b8e1cc8224c3b9602657444c93R66

Having a fixed or read-only repositories confines you to having that specific repository in that specific state. Any modifications done to the repository during a test, like deleting a branch, has to be undone for the next test otherwise they will fail. As tests need to expand to cover more areas, it will cause the read-only repository to become huge and bloated as more and more has to be added to it.
I believe this is how jgit does it's testing too.

wbrawner:
[...]

PR #1309, opened 2019-05-11
External repos zipped

chirontt:


fzs:
Thanks for having a go at this. I haven't tried it yet but looks good.
How large are the ambition and gitective repositories?

Seeing that the JGit repo is still 19MB after you took the time to clean it and to zip it, and checking where and how it is used in tests, I would rather not use it at all.

I want to merge your changes as they fix many test failures, especially the hello-world repository. So this helps right now.

In the long run I don't think the tests should use arbitrary repos from the internet but that we have synthesized repositories which have only the content needed for the various tests.

So while we can go with the other two for the time being, as a starting point to replace them with proper ones at some point, if they are small enough, I feel a bit uneasy assing 19MB binary weight to the project for only a few tests. So I would rather leave that out and either rewrite the tests or disable them for now.

rnveach:
Just my opinion, but as stated at https://github.com/gitblit/gitblit/issues/1275#issuecomment-373041192 , I recommend creating repositories at runtime.
Example: rnveach@d4b7fec#diff-40dc38b8e1cc8224c3b9602657444c93R82

fzs:
I think that there is merit in all three types of repositories for tests, prepared and checked in, created on the fly and external.
I do not believe in creating a repository on the fly for each and every test. I think that is too much overhead for many tests. For a lot of unit tests it is good to have one or two curated repositories put on disk that they can run against, especially if they only read from them. One curated repository can server a lot of tests.

For more complex tests or tests that require writing, putting together a repository during runtime makes sense, also in-memory. Again, maybe one repo can be used by multiple tests, so creating it might be part of a utility class instead of every single test.

And for a handful of integration tests external repositories will be needed, I think. But that is for the future.

chirontt:
My aim was to make the tests working again with _minimum_ changes to them. Rewriting them to use repos created on the fly is a very big job that I don't have time for. Maybe @rnveach's idea would be suitable for _new_ tests to be written, as we all know that more test cases are needed for the project, especially UI tests.

fzs:
I think zipping the JGit repository and adding 17MB is a bit much for the handful of tests that it is used for. I have therefore taken the JGit repository out of your commit. I have rewritten the commit, instead of adding a new one, so that the zip file doesn't become part of the history, taking up space.

So your fixes are in, but the JGit repository is still pulled from the Internet. This makes most tests pass again. At a later time the tests using the JGit repository can be rewritten to use a different repo.

I'm closing this pull request, as the changes are merged with commit e57a846.

Issue #938, opened 2015-10-07
All users were lost after a server failure (happend 2nd time to us)


igormukhin:
We had a server failure yesterday night (Blue screen, unexpected restart.) After the server and GitBlit (version 1.6.2) started again we found out that all our users and their repo permissions are gone. The file `users.conf` was completely new. Now we need to restore it from the backup.

This happen to us the second time. The first time it was 2 months ago. We had GitBlit 1.3.x or so.

It looks like GitBlit corrupts or deletes `users.conf` if it was stopped improperly.

I found no entry in the logs of GitBlit regarding the failure, so I looks like it was a complete failure of the operating system (BlueScreen).

This is what I found in the logs file at the server start:

```
2015-10-07 01:32:02 [ERROR] Failed to read D:\gitblit-data\users.conf
org.eclipse.jgit.errors.ConfigInvalidException: Cannot read file D:\gitblit-data\users.conf
	at org.eclipse.jgit.storage.file.FileBasedConfig.load(FileBasedConfig.java:173)
	at com.gitblit.ConfigUserService.read(ConfigUserService.java:884)
	at com.gitblit.ConfigUserService.getUserModel(ConfigUserService.java:189)
	at com.gitblit.manager.UserManager.getUserModel(UserManager.java:204)
	at com.gitblit.manager.AuthenticationManager.authenticate(AuthenticationManager.java:370)
	at com.gitblit.manager.AuthenticationManager.authenticate(AuthenticationManager.java:277)
	at com.gitblit.servlet.GitFilter.getUser(GitFilter.java:124)
	at com.gitblit.servlet.AccessRestrictionFilter.doFilter(AccessRestrictionFilter.java:159)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1650)
	at com.gitblit.servlet.ProxyFilter$1.doFilter(ProxyFilter.java:74)
	at com.gitblit.servlet.ProxyFilter.doFilter(ProxyFilter.java:77)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1650)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:583)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:143)
	at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:577)
[...]

PR #1364, opened 2021-03-20
Fix CPU hog in config save for large users.conf

Curly060:
If the realm file (users.conf) exceeds a certain size (about 30000 lines) saving this file causes a massive CPU hog. I have provided a branch that demonstrates the issue:
https://github.com/Curly060/gitblit/tree/show-user-config-save-cpu-hog-bug
https://github.com/gitblit/gitblit/compare/master...Curly060:show-user-config-save-cpu-hog-bug

So simply check out this branch and run the ConfigUserService class from there. But beware: Depending on the CPU you have, this might take from 20s up to several minutes!

This patch here simply replaces the usage of JGits Config classes with a much more efficient and far more simple class.

I know that Gitblit was not made for big workgroups and clearly this is a big users.conf. Nevertheless, saving 60000 lines should not hog a CPU like that.

### DIFF
--- a/src/main/java/com/gitblit/ConfigUserService.java
+++ b/src/main/java/com/gitblit/ConfigUserService.java
@@ -676,7 +676,7 @@ public class ConfigUserService implements IUserService {
 		// Write a temporary copy of the users file
 		File realmFileCopy = new File(realmFile.getAbsolutePath() + ".tmp");
 
-		StoredConfig config = new FileBasedConfig(realmFileCopy, FS.detect());
+		StoredUserConfig config = new StoredUserConfig(realmFileCopy);
 
 		// write users
 		for (UserModel model : users.values()) {
--- a/src/main/java/com/gitblit/StoredUserConfig.java
+++ b/src/main/java/com/gitblit/StoredUserConfig.java
@@ -1,0 +1,172 @@
+package com.gitblit;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.function.Function;
+
+/**
+ * Simple class with the only purpose to save the realm file (users.conf) in
+ * a fast efficient manner. The JGit Config classes used previously caused
+ * a massive CPU hog if the users file got bigger than about 30000 lines.
+ *
+ * @author Ingo Lafrenz
+ *
+ */
+public class StoredUserConfig {
+
+	private final File realmFileCopy;
+	private SortedMap<String, Section> sections = new TreeMap<>();
+
+	public StoredUserConfig(File realmFileCopy) {
+		this.realmFileCopy = realmFileCopy;
+	}
+
+	public void setString(final String section, final String subsection, String name, String value) {
+		Section s = sections.computeIfAbsent(generateKey(section, subsection), new Function<String, Section>() {
+			@Override
+			public Section apply(String k) {
+				return new Section(section, subsection);
+			}
+		});
+		s.addEntry(name, value);
+	}
+
+	public void setBoolean(String section, String subsection, String name, boolean value) {
+		setString(section, subsection, name, String.valueOf(value));
+	}
+
+	public void setStringList(String section, String subsection, String name, List<String> list) {
+		for (String value : list) {
+			setString(section, subsection, name, value);
+		}
+	}
+
+	public void save() throws IOException {
+		try (FileWriter fileWriter = new FileWriter(realmFileCopy);
+				PrintWriter printWriter = new PrintWriter(fileWriter);) {
+			for (Map.Entry<String,Section> entry : sections.entrySet()) {
+				writeSection(printWriter, entry.getKey(), entry.getValue());
+			}
+		}
+	}
+
+	private static void writeSection(PrintWriter printWriter, String key, Section section) {
+		printWriter.printf("[%s \"%s\"]\n", section.getName(), section.getSubSection());
+		for (Entry entry : section.getEntries().values()) {
+			writeEntry(printWriter, entry.getKey(), entry.getValue());
+		}
+	}
+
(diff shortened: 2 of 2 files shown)


flaix:
Hi Ingo!

Thanks you for bringing this to attention and providing a fix. I will have a look at it as soon as I find the time.
Out of curiosity, do you run Gitblit for such large workgroups or how did you encounter that behaviour?

Curly060:
Gitblit is used at the company I work at and this company is indeed a very big global company. We use a forked version of the Gitblit Atlassian Crowd Plugin which syncs users from Crowd to Gitblit. Unfortunately the plugin only ever adds new users and never removes dead ones, so over time the users.conf has grown quite large (~62000 lines).
So recently gitblit had started to hang for several minutes quite often. Eventually this got so frustrating that I finally investigated the problem in detail.
tl;dr: JGit internally uses an array to handle Config files and about 1 billion(!!) loop runs occur when storing those 62k lines.

Now the gory details ;)
When Gitblit saves the users.conf, it rebuilds the config from scratch like this:
for (user : users) {
  config.setXXX(); // with up to 18 such setXXX() methods
}
for (team : teams) {
  config.setXXX(); //with up to 8 such setXXX() methods
}
These methods all internally call the setStringList() method which calls replaceStringList(). The JGit Config class  holds an array of ConfigLine items.  And for every setXXX method it first creates a new array of size "old array size + string list size",copies all entries and then always(!) runs through that entire(!) array doing some string comparisons ("am I in the right section and subsection?").
While this may be fine for small config files like they exist in typical git repositories, it certainly is not when the files grow larger. As I was saying, 62k lines result in approx. 1 billion internal loop runs in the Config class.

What makes the whole situation even worse is that the write method of ConfigUserService is synchronized! So if this method takes a minute, all other attempts to write need to wait this minute. And the next write of course takes another minute again. It's quite deadly ;)

So that's why I decided to replace the config save logic with a much simpler one. Now saving takes 200ms or so.

flaix:
Great!

Have you tested your class and could you provide some unit tests for it? I am trying to increase the test coverage to make supporting Gitblit easier in the future.
Have you ever encountered something like in #938?  Are you sure that your class is not susceptible to data loss?

That sounds like an impressive setup. If you have forked the Crowd plugin, would it be easy to make it also delete users? With a quick search I found two plugins, but didn't have a closer look.
Can I ask how large your developer base in Gitblit is, i.e. actual developers, not dead user entries? If you have many developers maybe there are other problems that you have come across with Gitblit?

Curly060:
I have made the following tests, that were all successful:

- Load existing config file and write it again with both, old and new logic, then sort and compare => both files were identical
- run the patched version of Gitblit in our test environment (with Crowd plugin enabled) and make operations that trigger saving the users.conf
- run a local patched version against some of my own repositories with some local tests users and make operations that trigger saving the users.conf

But you are quite right, this may not be enough testing. After all, our users do not have all properties of the UserModel set. So yes, unit tests are probably a very good idea. However, I am not that deep in Gitblit source that I may catch all cases - and I will have to find the time to write those. So this might take a while.

About issue https://github.com/gitblit/gitblit/issues/938:
I wouldn't expect my class to cause such behaviour. It simply replaces StoredConfig / FileBasedConfig of JGit with a different implementation. The rest of the logic of GitBlit stays the same: build whole config first, then save (to a temp file) and only if that temp file exists and is not empty, do the rename.

About the crowd plugin: I haven't had a look at the code there yet, but yes, it should be possible to delete users again. However, fixing the config saving seemed like easiest thing to do :)

About the developer base: Hard to say, but in the last 7 days there were commits from ~180 different committers. Using git shortlog and some bash scripting: ~380 commiters in the past year. Amount of repos: 660

So far, Gitblit has served well for the past couple of years, I'd say. This config save thing was the only major issue that I ways aware of in the past 3 years that I work there.

Curly060:
I just noticed: The UserServiceTest class already implicitly tests pretty much everything. It creates various attributes in UserModel and TeamModel, saves them to the file individually, reads the file again and checks if those attributes are still the same. Would that not be sufficient testing then?

flaix:
Thanks for your answer and for enduring my questions. :)
Yeah, I guess you are right that the existing tests should suffice.

martinspielmann:
> Have you tested your class and could you provide some unit tests for it? I am trying to increase the test coverage to make supporting Gitblit easier in the future.
> Have you ever encountered something like in #938? Are you sure that your class is not susceptible to data loss?

I'm still sorry about this one :cry: 
Really nice catch and cool implementation, Ingo!

martinspielmann:
Really nice catch and cool implementation, Ingo!

Curly060:
We have built gitblit with this patch and deployed it on prod a week ago. This has solved the CPU hogging problem completely and now gitblit runs just fine.

Curly060:
This change has now been active for over 3 months in our company and it runs perfectly fine and without any issues. Any chance this gets merged?

flaix:
Ya, I had an issue with it, why it didn't use the `StoredConfig` interface, but since I haven't gotten around to comment or change it since then, and since it is proven to work and fixes a problem, I just merged right now.

Curly060:
About the `StoredConfig`:
I have done this on purpose to clearly show that in this place using the JGit Config class is not applicable at all!
The trouble with `StoredConfig` is that it unfortunately is **not** really an interface, it's an abstract class extending the  `Config` class from JGit.
If I had extended from `StoredConfig` not only this would have forced me to override `load()` (which makes no sense in my implementation), no, it would have pulled in all the `setXXX` methods which are responsible for this CPU hog in the first place.
But the whole point of this fix was to avoid exactly these methods, hence, a completely new implementation.

Yes, I know this is not ideal, but actually it is the fault of JGit which does not provide clean interfaces for its `Config` classes...

Oh, and thanks a lot for merging :)

flaix:
Ah, yes, that rings a bell. I guess I had found that out, too, but then for some reason forgot about the PR again. So thanks for the reminder.

flaix:
Or, checking the builds, maybe it was because it uses Java 8, so it would need to be part of 1.10, as I keep 1.9 on Java 7.

PR #1370, opened 2021-05-26
Modified Batch file for Windows, if it's called directly

Zwixx:
I was tricked by the first install and clicked on my Windows 10 on the gitblit.cmd and nothing happens. I looked into it and a environment Variable "CD" is used, but never set. I modified the batch file, so the current directory of the batch file is used as root folder for the classpath. So it worked without anything else to modify.

So nothing like set up a environment variable is needed.

### DIFF
--- a/src/main/distrib/win/gitblit-stop.cmd
+++ b/src/main/distrib/win/gitblit-stop.cmd
@@ -2,4 +2,4 @@
 
 SET CD=%~dp0
 SET CD=%CD:~0,-1%
-java -cp gitblit.jar;"%CD%\ext\*" com.gitblit.GitBlitServer --stop --baseFolder data %*
+java -cp "%CD%\gitblit.jar";"%CD%\ext\*" com.gitblit.GitBlitServer --stop --baseFolder data %*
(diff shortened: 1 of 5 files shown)


flaix:
The CD variable is a dynamic variable and does not need to be set explicitly. It is automatically set by cmd.exe. 
Did you encounter problems because you do call the batch file from a different directory than the Gitblit directory, or because you somehow have CMD Extensions disabled?

In any case I would prefer to use a dedicated variable, like `gbhome` or something like that, instead of overwriting the automatic variable.

Zwixx:
I automated it with the build-in variables from Windows NT cmd. It is the current directory, but not the directory of the batch file.
So it don't works if you start it for example in the Windows explorer.

flaix:
Yes, it wouldn't have worked for that use case. Thank you for fixing it. If you don't mind, I'll change it to use a different variable name.

flaix:
Actually, I cannot reproduce this problem. Testing this under Windows 10 when I double-click on `gitblit.cmd` in the Windows Explorer, the server starts without problems. It even works when I create a link and place it in a different directory, then double-click the link.

What does not work, is running the `gitblit.cmd` command on the command line from a different folder, e.g. `gitblit 1.9.1\gitblit.cmd`. But this would even need one more change, setting the base folder correctly.

I don't know if the latter is a often used use case, but we can still make it work. Still, I'll rather do it without setting explicit variables.

PR #1438, opened 2022-11-20
Replace SWF clippy with clipboardjs on repository page

flaix:
Replace the outdated and broken Shockwave Flash way of copying text to the browser clip board with a modern JavaScript approach. This introduces the JS library [clipboardjs](https://clipboardjs.com).

There is some visual feedback in form of a tooltip. While this works fine for the main repository URL, it is less so for the copy buttons in the drop-down menus. The tool tip is attached to a element higher in the hierarchy so that it gets shown. But the height seems to be restricted, so the tooltip doesn't have the proper padding.


flaix:
Has been merged. Status closed is incorrect.

